PR #30, opened 2021-04-01
Add splunkhttp.NewHandler

pellared:
## Why 

https://github.com/signalfx/splunk-otel-go/pull/18#issuecomment-808338996

>If we could add some form of this to the distro package. That would seem to make sense to me given it is supposed to be the super mega deluxe easy path for setup Splunk customers.

## What

- Add `splunkhttp.NewHandler`

## Next

- https://github.com/signalfx/splunk-otel-go/pull/30#discussion_r606000990

### DIFF
--- a/instrumentation/net/http/splunkhttp/config.go
+++ b/instrumentation/net/http/splunkhttp/config.go
@@ -34,6 +34,11 @@ func WithOTelOpts(opts ...otelhttp.Option) Option {
 }
 
 // WithServerTiming enables or disables the ServerTimingMiddleware.
+//
+// The default is to enable the ServerTimingMiddleware if this option is not passed.
+// Additionally, the SPLUNK_CONTEXT_SERVER_TIMING_ENABLED environment variable
+// can be set to TRUE or FALSE to specify this option. This option value will be
+// given precedence if both it and the environment variable are set.
 func WithServerTiming(v bool) Option {
 	return optionFunc(func(cfg *config) {
 		cfg.ServerTimingEnabled = v
--- a/instrumentation/net/http/splunkhttp/config.go
+++ b/instrumentation/net/http/splunkhttp/config.go
@@ -1,0 +1,76 @@
+// Copyright Splunk Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package splunkhttp
+
+import (
+	"os"
+	"strings"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+// Environmental variables used for configuration.
+const (
+	EnvVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED" // Adds `Server-Timing` header to HTTP responses
+)
+
+// WithOTelOpts is use to pass the OpenTelemetry SDK options.
+func WithOTelOpts(opts ...otelhttp.Option) Option {
+	return optionFunc(func(cfg *config) {
+		cfg.OTelOpts = opts
+	})
+}
+
+// WithServerTiming enabled or disabled the ServerTimingMiddleware.
+func WithServerTiming(v bool) Option {
+	return optionFunc(func(cfg *config) {
+		cfg.ServerTimingEnabled = v
+	})
+}
+
+// Option is used for setting *optional* config properties.
+type Option interface {
+	apply(*config)
+}
+
+// config represents the available configuration options.
+type config struct {
+	OTelOpts            []otelhttp.Option
+	ServerTimingEnabled bool
+}
+
+// optionFunc provides a convenience wrapper for simple Options
+// that can be represented as functions.
+type optionFunc func(*config)
+
+func (o optionFunc) apply(c *config) {
+	o(c)
+}
+
+// newConfig creates a new config struct and applies opts to it.
+func newConfig(opts ...Option) *config {
+	serverTimingEnabled := true
+	if v := os.Getenv(EnvVarServerTimingEnabled); strings.EqualFold(v, "false") {
+		serverTimingEnabled = false
+	}
+
+	c := &config{
+		ServerTimingEnabled: serverTimingEnabled,
+	}
+	for _, opt := range opts {
+		opt.apply(c)
+	}
+	return c
+}
--- a/instrumentation/net/http/splunkhttp/config.go
+++ b/instrumentation/net/http/splunkhttp/config.go
@@ -23,7 +23,7 @@ import (
 
 // Environmental variables used for configuration.
 const (
-	EnvVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED" // Adds `Server-Timing` header to HTTP responses
+	envVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED" // Adds `Server-Timing` header to HTTP responses
 )
 
 // WithOTelOpts is use to pass the OpenTelemetry SDK options.
@@ -33,14 +33,14 @@ func WithOTelOpts(opts ...otelhttp.Option) Option {
 	})
 }
 
-// WithServerTiming enabled or disabled the ServerTimingMiddleware.
+// WithServerTiming enables or disables the ServerTimingMiddleware.
 func WithServerTiming(v bool) Option {
 	return optionFunc(func(cfg *config) {
 		cfg.ServerTimingEnabled = v
 	})
 }
 
-// Option is used for setting *optional* config properties.
+// Option is used for setting optional config properties.
 type Option interface {
 	apply(*config)
 }
@@ -62,7 +62,7 @@ func (o optionFunc) apply(c *config) {
 // newConfig creates a new config struct and applies opts to it.
 func newConfig(opts ...Option) *config {
 	serverTimingEnabled := true
-	if v := os.Getenv(EnvVarServerTimingEnabled); strings.EqualFold(v, "false") {
+	if v := os.Getenv(envVarServerTimingEnabled); strings.EqualFold(v, "false") {
 		serverTimingEnabled = false
 	}
 
--- a/instrumentation/net/http/splunkhttp/config_test.go
+++ b/instrumentation/net/http/splunkhttp/config_test.go
@@ -28,7 +28,7 @@ func TestConfigs(t *testing.T) {
 		envs   map[string]string
 		assert func(t *testing.T, c *config)
 	}{
-		// Defaiult
+		// Default
 		{
 			name: "Default",
 			assert: func(t *testing.T, c *config) {
--- a/instrumentation/net/http/splunkhttp/config_test.go
+++ b/instrumentation/net/http/splunkhttp/config_test.go
@@ -1,0 +1,89 @@
+// Copyright Splunk Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package splunkhttp
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConfigs(t *testing.T) {
+	tests := []struct {
+		name   string
+		opts   []Option
+		envs   map[string]string
+		assert func(t *testing.T, c *config)
+	}{
+		// Defaiult
+		{
+			name: "Default",
+			assert: func(t *testing.T, c *config) {
+				assert.Nil(t, c.OTelOpts, "should not set any additional OTel options")
+				assert.True(t, c.ServerTimingEnabled, "should enable ServerTiming")
+			},
+		},
+		// ServerTiming
+		{
+			name: "ServerTiming WithServerTiming(false)",
+			opts: []Option{
+				WithServerTiming(false),
+			},
+			assert: func(t *testing.T, c *config) {
+				assert.False(t, c.ServerTimingEnabled, "should disable ServerTiming")
+			},
+		},
+		{
+			name: "ServerTiming SPLUNK_CONTEXT_SERVER_TIMING_ENABLED=False",
+			envs: map[string]string{
+				"SPLUNK_CONTEXT_SERVER_TIMING_ENABLED": "False",
+			},
+			assert: func(t *testing.T, c *config) {
+				assert.False(t, c.ServerTimingEnabled, "should disable ServerTiming")
+			},
+		},
+		{
+			name: "ServerTiming WithServerTiming(true) SPLUNK_CONTEXT_SERVER_TIMING_ENABLED=True",
+			envs: map[string]string{
+				"SPLUNK_CONTEXT_SERVER_TIMING_ENABLED": "False",
+			},
+			opts: []Option{
+				WithServerTiming(true),
+			},
+			assert: func(t *testing.T, c *config) {
+				assert.True(t, c.ServerTimingEnabled, "should enable ServerTiming, because option has higher priority than env var")
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			// set env var before the test and bring the old values back after the test
+			for key, val := range tt.envs {
+				if v, ok := os.LookupEnv(key); ok {
+					defer func() { os.Setenv(key, v) }()
+				} else {
+					defer func() { os.Unsetenv(key) }()
+				}
+				os.Setenv(key, val)
+			}
+
+			cfg := newConfig(tt.opts...)
+			tt.assert(t, cfg)
+		})
+	}
+}
--- a/instrumentation/net/http/splunkhttp/handler.go
+++ b/instrumentation/net/http/splunkhttp/handler.go
@@ -22,8 +22,9 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
+// Environmental variables used for configuration.
 const (
-	EnvVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED"
+	EnvVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED" // Adds `Server-Timing` header to HTTP responses
 )
 
 // NewHandler wraps the passed handler in a span named after the operation and with any provided Options.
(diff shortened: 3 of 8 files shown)


instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+// WithOtelOpts is use to pass the OpenTelemetry SDK options.
+func WithOtelOpts(opts ...otelhttp.Option) Option {
+	return OptionFunc(func(cfg *config) {
+		cfg.OtelOpts = opts
+	})
+}
+
+// WithServerTiming enabled or disabled the ServerTimingMiddleware.
+func WithServerTiming(v bool) Option {
```
owais, line 50: How about `WithoutServerTiming()` or `WithServerTimingDisabled()`? This wouldn't take any parameters and always set the field to `false`.

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Option Interface used for setting *optional* config properties
+type Option interface {
+	Apply(*config)
+}
+
+// config represents the available configuration options.
+type config struct {
+	OtelOpts            []otelhttp.Option
+	ServerTimingEnabled bool
```
owais, line 64: do we want to keep these options as public considering we have the `With...` helper methods. It's easier to make them public later but impossible to make them private without major version change :)

pellared: Struct is not exported so it is not public API

instrumentation/net/http/splunkhttp/handler.go
pellared: It was my initial design. However I feel that current is more flexible if we would like to change the default value or the client would like to get the value from CLI arg using flag package

instrumentation/net/http/splunkhttp/handler.go
owais: So why not keep fields lower case as well at least for self-documentation?

pellared: If we would like to export it at some point then it would be easier 😉 
It is also consistent with:
- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/2760096474280b3634e92431a7ed6da0744ef144/instrumentation/net/http/otelhttp/config.go#L34-L46
- https://github.com/signalfx/splunk-otel-go/blob/59c68c77a30c9c73fff5a8edcb70e4548b7535a2/distro/config.go#L34-L38
It is for me like having `public` methods in `internal` classes in .NET or `package` in Java.

@MrAlias What do you think?

pellared:
@MrAlias Any idea what is going on with MacOS runners? https://github.com/signalfx/splunk-otel-go/pull/30/checks?check_run_id=2246124758 (I cannot investigate it today - I can look tomorrow)

instrumentation/net/http/splunkhttp/handler.go
owais: We can just update them if we want to expose this later. Imagine another dev in future exposing config struct and then unintentionally exposing all the fields because they were capitalized.

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+const (
+	EnvVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED"
+)
+
+// NewHandler wraps the passed handler, functioning like Splunk specific http middleware middleware,
+// in a span named after the operation and with any provided Options.
```
MrAlias, line 30:
```suggestion
// NewHandler wraps the passed handler in a span named after the operation and with any provided Options.
// This will also enable all the Splunk specific defaults for HTTP tracing.
```

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+	if cfg.ServerTimingEnabled {
+		handler = ServerTimingMiddleware(handler)
+	}
+
+	handler = otelhttp.NewHandler(handler, "server", cfg.OtelOpts...)
+	return handler
+}
+
+// WithOtelOpts is use to pass the OpenTelemetry SDK options.
+func WithOtelOpts(opts ...otelhttp.Option) Option {
```
MrAlias, line 43:
```suggestion
func WithOTelOpts(opts ...otelhttp.Option) Option {
```

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+// NewHandler wraps the passed handler, functioning like Splunk specific http middleware middleware,
+// in a span named after the operation and with any provided Options.
+func NewHandler(handler http.Handler, operation string, opts ...Option) http.Handler {
+	cfg := newConfig(opts...)
+
+	if cfg.ServerTimingEnabled {
+		handler = ServerTimingMiddleware(handler)
+	}
+
+	handler = otelhttp.NewHandler(handler, "server", cfg.OtelOpts...)
```
MrAlias, line 38: `operation` is not passed to the `NewHandler`, instead a static `"server"` is used. Is this intentional?

instrumentation/net/http/splunkhttp/handler.go
MrAlias: It doesn't make too much of a difference at this point as @pellared pointed out.

One nice thing to also mention is that if we continue to unexport `config` but want to unmarshal into it in the future (i.e. yaml or json configuration) these fields will need to be exported.

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+// config represents the available configuration options.
+type config struct {
+	OtelOpts            []otelhttp.Option
+	ServerTimingEnabled bool
+}
+
+// OptionFunc provides a convenience wrapper for simple Options
+// that can be represented as functions.
+type OptionFunc func(*config)
```
MrAlias, line 69:
```suggestion
type optionFunc func(*config)
```

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (o OptionFunc) Apply(c *config) {
+	o(c)
+}
+
+// newConfig creates a new config struct and applies opts to it.
+func newConfig(opts ...Option) *config {
+	serverTimingEnabled := true
+	if v := os.Getenv(EnvVarServerTimingEnabled); v == "0" || strings.EqualFold(v, "false") {
```
MrAlias, line 78: `"0"` doesn't seem to be a valid value according to the specification: https://github.com/signalfx/gdi-specification/blob/main/docs/configuration.md#instrumentation-libraries

Should this just check for `false`?

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Option Interface used for setting *optional* config properties
+type Option interface {
+	Apply(*config)
+}
+
+// config represents the available configuration options.
+type config struct {
+	OtelOpts            []otelhttp.Option
```
MrAlias, line 63:
```suggestion
	OTelOpts            []otelhttp.Option
```

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+	serverTimingEnabled := true
+	if v := os.Getenv(EnvVarServerTimingEnabled); v == "0" || strings.EqualFold(v, "false") {
+		serverTimingEnabled = false
+	}
+
+	c := &config{
+		ServerTimingEnabled: serverTimingEnabled,
+	}
+	for _, opt := range opts {
+		opt.Apply(c)
```
MrAlias, line 86: The precedence of passed configuration overriding environment variables should be tested.

Maybe something similar to [this](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlphttp/options_test.go).

instrumentation/net/http/splunkhttp/server_timing_test.go
```diff
@@ -26,21 +25,44 @@ import (
...
+		handler = ServerTimingMiddleware(handler)
+		return otelhttp.NewHandler(handler, "server", otelhttp.WithTracerProvider(oteltest.NewTracerProvider()))
+	})
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode, "should return OK status code")
+	assert.Contains(t, resp.Header["Access-Control-Expose-Headers"], "Server-Timing", "should set Access-Control-Expose-Headers header")
+	assert.Regexp(t, "^traceparent;desc=\"00-[0-9a-f]{32}-[0-9a-f]{16}-01\"$", resp.Header.Get("Server-Timing"), "should return properly formated Server-Timing header")
+}
+
+func TestNewHandler_Default(t *testing.T) {
```
MrAlias, line 38:
```suggestion
func TestNewHandlerDefault(t *testing.T) {
```

instrumentation/net/http/splunkhttp/server_timing_test.go
```diff
@@ -26,21 +25,44 @@ import (
...
+	resp := responseForHandler(func(handler http.Handler) http.Handler { // nolint:bodyclose // Body is not used
+		return NewHandler(handler, "server", WithOtelOpts(otelhttp.WithTracerProvider(oteltest.NewTracerProvider())))
+	})
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode, "should return OK status code")
+	assert.Contains(t, resp.Header["Access-Control-Expose-Headers"], "Server-Timing", "should set Access-Control-Expose-Headers header")
+	assert.Regexp(t, "^traceparent;desc=\"00-[0-9a-f]{32}-[0-9a-f]{16}-01\"$", resp.Header.Get("Server-Timing"), "should return properly formated Server-Timing header")
+}
+
+func TestNewHandler_ServerTimingDisabled(t *testing.T) {
```
MrAlias, line 48:
```suggestion
func TestNewHandlerServerTimingDisabled(t *testing.T) {
```

instrumentation/net/http/splunkhttp/handler.go
MrAlias, line 43: This feels like it is going to require a ton of boilerplate to configure things. I wonder if upstream, the `config` should be exported so we could embed it here and just have our options extend upstream. This isn't something blocking this PR, but more so wondering what others thoughts are.

instrumentation/net/http/splunkhttp/handler.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"net/http"
+	"os"
+	"strings"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+const (
+	EnvVarServerTimingEnabled = "SPLUNK_CONTEXT_SERVER_TIMING_ENABLED"
```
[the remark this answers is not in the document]
pellared: I thought it may be good to export. Env vars are kind of public API anyway. Someone may like to use it 😉 I should probably add a comment anyway.

instrumentation/net/http/splunkhttp/handler.go
pellared: It is a 🐛 . Nice catch 👍

instrumentation/net/http/splunkhttp/handler.go
pellared: I believe I saw this pattern of accepting `1` as `true` and `0` as `false` somewhere... But I agree with you. I checked the code and this is e.g how Java parsing is done: via [`Boolean::parseBoolean()`](https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html)

Will fix that

instrumentation/net/http/splunkhttp/handler.go
pellared: will do

instrumentation/net/http/splunkhttp/server_timing_test.go
pellared: Will change even though my eyes would even prefer `Test_NewHandler_default` 😄 But I do not have a strong opinion 😉

instrumentation/net/http/splunkhttp/handler.go
pellared: For me it makes sense. I think linters could complain in line :https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/config.go#L50 as I feel it makes the API less extensible. I can try tackling it after this PR is merged.

instrumentation/net/http/splunkhttp/handler.go
pellared: I also unexported `apply` in `Option` type to:
```
// Option is used for setting *optional* config properties.
type Option interface {
	apply(*config)
}
```

instrumentation/net/http/splunkhttp/handler.go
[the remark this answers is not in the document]
pellared: Added comments https://github.com/signalfx/splunk-otel-go/pull/30/commits/33dec16e4a16090a3f0b13b567f4f30593df4daf

instrumentation/net/http/splunkhttp/handler.go
pellared: Addressed https://github.com/signalfx/splunk-otel-go/pull/30/commits/33dec16e4a16090a3f0b13b567f4f30593df4daf

instrumentation/net/http/splunkhttp/handler.go
pellared: Addressed https://github.com/signalfx/splunk-otel-go/pull/30/commits/33dec16e4a16090a3f0b13b567f4f30593df4daf

instrumentation/net/http/splunkhttp/handler.go
pellared: Addressed https://github.com/signalfx/splunk-otel-go/pull/30/commits/33dec16e4a16090a3f0b13b567f4f30593df4daf

instrumentation/net/http/splunkhttp/server_timing_test.go
pellared: Addressed https://github.com/signalfx/splunk-otel-go/pull/30/commits/33dec16e4a16090a3f0b13b567f4f30593df4daf

instrumentation/net/http/splunkhttp/handler.go
pellared: Addressed https://github.com/signalfx/splunk-otel-go/pull/30/commits/33dec16e4a16090a3f0b13b567f4f30593df4daf

instrumentation/net/http/splunkhttp/handler.go
pellared: @MrAlias I just want to let you know that the tests you have send are not really testing the precedence correctly b/c the precedence logic is in test code: https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlphttp/options_test.go#L289-L292

pellared: Addressed https://github.com/signalfx/splunk-otel-go/pull/30/commits/076f759d6276bbf0b37439611cfcaa420439f180

instrumentation/net/http/splunkhttp/server_timing_test.go
pellared: Addressed 33dec16

instrumentation/net/http/splunkhttp/handler.go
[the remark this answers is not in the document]
MrAlias: I'm pretty hesitant to export this at this point. It will set a precedent and I'm not sure it is one we will or won't want in the long term.

I'm not aware of a situation where a user will need to reference this variable given it is expected to describe something set in the runtime environment prior to execution. But, if we do not export this now and there arises a need the future it will be simple enough to export it. In that situation we will probably want to look holistically at exporting all environment variables Splunk uses, likely in some kind of "splunk environment variable" package. If we export this now it will either require us to duplicate the declaration in both packages or be a breaking change.

instrumentation/net/http/splunkhttp/config.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// WithOTelOpts is use to pass the OpenTelemetry SDK options.
+func WithOTelOpts(opts ...otelhttp.Option) Option {
+	return optionFunc(func(cfg *config) {
+		cfg.OTelOpts = opts
+	})
+}
+
+// WithServerTiming enabled or disabled the ServerTimingMiddleware.
```
MrAlias, line 36:
```suggestion
// WithServerTiming enables or disables the ServerTimingMiddleware.
//
// The default is to enable the ServerTimingMiddleware if this option is not passed.
// Additionally, the SPLUNK_CONTEXT_SERVER_TIMING_ENABLED environment
// variable can be set to TRUE or FALSE to specify this option. This option value will
// be given precedence if both it and the environment variable are set.
```

instrumentation/net/http/splunkhttp/config.go
```diff
@@ -0,0 +1,76 @@
...
+func WithServerTiming(v bool) Option {
+	return optionFunc(func(cfg *config) {
+		cfg.ServerTimingEnabled = v
+	})
+}
+
+// Option is used for setting *optional* config properties.
```
MrAlias, line 43: Go docs does not interpret this as markdown

```suggestion
// Option is used for setting optional config properties.
```

instrumentation/net/http/splunkhttp/config_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func TestConfigs(t *testing.T) {
+	tests := []struct {
+		name   string
+		opts   []Option
+		envs   map[string]string
+		assert func(t *testing.T, c *config)
+	}{
+		// Defaiult
```
MrAlias, line 31:
```suggestion
		// Default
```

MrAlias (commented):
minor doc edits

instrumentation/net/http/splunkhttp/handler.go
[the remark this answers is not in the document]
pellared: Unexported https://github.com/signalfx/splunk-otel-go/pull/30/commits/165c57c14f22bd11afdca17c60347dcc4b9c7208

instrumentation/net/http/splunkhttp/config.go
pellared: Fixed https://github.com/signalfx/splunk-otel-go/pull/30/commits/165c57c14f22bd11afdca17c60347dcc4b9c7208

instrumentation/net/http/splunkhttp/config_test.go
pellared: Fixed https://github.com/signalfx/splunk-otel-go/pull/30/commits/165c57c14f22bd11afdca17c60347dcc4b9c7208

instrumentation/net/http/splunkhttp/config.go
MrAlias: Did you not want to add the paragraph about defaults and overrides with the environment variable?

pellared: You mean https://github.com/signalfx/splunk-otel-go/blob/165c57c14f22bd11afdca17c60347dcc4b9c7208/instrumentation/net/http/splunkhttp/README.md ?

MrAlias: Hmm, not sure I follow.

I had originally suggested the `s/(en|dis)abled/\1ables/` change which looks a part of [165c57c](https://github.com/signalfx/splunk-otel-go/commit/165c57c14f22bd11afdca17c60347dcc4b9c7208), but also suggested we add the following as well to the `WithServerTiming` documentation:

```
// The default is to enable the ServerTimingMiddleware if this option is not passed.
// Additionally, the SPLUNK_CONTEXT_SERVER_TIMING_ENABLED environment
// variable can be set to TRUE or FALSE to specify this option. This option value will
// be given precedence if both it and the environment variable are set.
```

pellared: It is me how was not following. Give me a minute, please and sorry for inconvenience  😉

pellared: Added here: https://github.com/signalfx/splunk-otel-go/pull/30/commits/092496b8b3a7ff804173dc420f8fb27e1b145cb9

MrAlias (approved):

MrAlias:
@owais PTAL

MrAlias (approved):

owais (approved):

PR #33, opened 2021-04-09
[PoC] Refactor config API

pellared:
## Why

https://github.com/signalfx/splunk-otel-go/pull/30#discussion_r606000990

## What

See: https://github.com/signalfx/splunk-otel-go/pull/33/commits/2c907eb8cc7c3698014c21c151ea5288c59be0f7 and https://github.com/signalfx/splunk-otel-go/pull/33/commits/d8365c1504837a5ce28cdb9ddd4746c4c1052964
It is an attempt to make the OTel functional options API more extensible-friendly.


instrumentation/net/http/otelhttp/config.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return c
+}
+
+// WithNop returns a no operation option.
+// It can be used for extensibilty.
+func WithNop() Option {
+	return OptionFunc(func(*config) {
+	})
+}
```
pellared, line 89: This is the only change in OTel SDK. This is a way that requires the fewest code changes in OTel SDK.

I think a better alternative may be to export the `config` type and at the same time unexport all fields in the `config` struct. It would give the possibility to implement the `Option` interface which right now was not really possible (without the `WithNop` hack).

Any other ideas?

pellared:
@MrAlias @owais 

I tried 3 approaches to make OTel functional options API more extensible-friendly.

1. Add `WithNop` functional option to make it easier to do a composition. It was done here: https://github.com/signalfx/splunk-otel-go/commit/2c907eb8cc7c3698014c21c151ea5288c59be0f7
2. Export `Config` in `otelhttp` and at the same time unexport all its fields.
3. Change `Apply` in `Option` from `otelhttp` to accept `interface{}` instead of `*config` (current PR state)

### Approach 1

I do not like adding the `WithNop` function to the public API surface. IMO the worst approach.

### Approach 2

It does not have the drawback of approach 1, however, it will result in a similar code boat to opinion 2, It requires type-assertions (but not in OTel code - just in "distro" code). _When I started writing this code I ended with approach no 3._

### Approach 3

I feel that this one is the cleanest. We lose some kind of type-safety as we really on type assertions, but I cannot figure out any better way that will work nicely with the functional options approach. The implementation of options in OTEL and distro is similar (orthogonal).


I was also thinking of builder pattern or simply using config structs... But after a few hours of experiments, I just feel that I am not able to come up with anything better than presented in this PR.

instrumentation/net/http/splunkhttp/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/signalfx/splunk-otel-go/instrumentation/net/http/splunkhttp
 
 go 1.14
 
+replace go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => ../otelhttp
```
pellared, line 5: I have "vendored" `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` module to easily experiment and visualize the changes in OTel repo

instrumentation/net/http/otelhttp/config.go
```diff
@@ -0,0 +1,175 @@
...
+
+// OptionFunc provides a convenience wrapper for simple Options
+// that can be represented as functions.
+type OptionFunc func(*config)
+
+func (o OptionFunc) Apply(obj interface{}) {
+	if c, ok := obj.(*config); ok {
+		o(c)
+	}
+}
```
pellared, line 61: these are the only changes required in OTel repo

pellared:
@MrAlias @owais Can you please take a look and comment?

MrAlias:
Awesome work! :smile: 

I like option 2 the most due to the type safety, but I'm still thinking it through. I definitely want to present this to the Go SIG on Thursday to get more feedback.

pellared:
> Awesome work! 😄
> 
> I like option 2 the most due to the type safety, but I'm still thinking it through. I definitely want to present this to the Go SIG on Thursday to get more feedback.

I will try to make a draft PR for each approach + GitHub issue in [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib) to facilitate the discussion.

pellared:
@MrAlias 

**Approach 1:** https://github.com/signalfx/splunk-otel-go/pull/37
**Approach 2:** https://github.com/signalfx/splunk-otel-go/pull/39
**Approach 3:** https://github.com/signalfx/splunk-otel-go/pull/38

PR #37, opened 2021-04-14
[PoC] Refactor config composition - WithNop approach

pellared:


instrumentation/net/http/otelhttp/config.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return c
+}
+
+// WithNop returns a no operation option.
+// It can be used for extensibilty.
+func WithNop() Option {
+	return OptionFunc(func(*config) {
+	})
+}
```
pellared, line 89: This is the only change in OTel SDK.

PR #38, opened 2021-04-14
[PoC] Refactor config composition - interface{} approach 

pellared:


instrumentation/net/http/otelhttp/config.go
```diff
@@ -0,0 +1,175 @@
...
+
+// OptionFunc provides a convenience wrapper for simple Options
+// that can be represented as functions.
+type OptionFunc func(*config)
+
+func (o OptionFunc) Apply(obj interface{}) {
+	if c, ok := obj.(*config); ok {
+		o(c)
+	}
+}
```
pellared, line 61: These are the only changes in OTel SDK.

PR #39, opened 2021-04-14
[PoC] Refactor config composition - exported Config approach 

pellared:


instrumentation/net/http/otelhttp/config.go
```diff
@@ -0,0 +1,173 @@
...
+	propagators       propagation.TextMapPropagator
+	spanStartOptions  []trace.SpanOption
+	readEvent         bool
+	writeEvent        bool
+	filters           []Filter
+	spanNameFormatter func(string, *http.Request) string
+
+	tracerProvider trace.TracerProvider
+	meterProvider  metric.MeterProvider
+}
```
pellared, line 46: the exporting is inverted here. 
all usages in this package have also changed accordingly

PR #40, opened 2021-04-15
Refactor splunkhttp config processing

pellared:
## Why

https://github.com/signalfx/splunk-otel-go/pull/30#discussion_r606000990

Make the functional options API more user-friendly. 
Currently, sample usage looks like this: 

```go
handler = splunkhttp.NewHandler(handler, "server", splunkhttp.WithOTelOpts(otelhttp.WithTracerProvider(provider)), splunkhttp.WithTraceResponseHeader(false))
```

## What

Refactor config processing in `splunkhttp` package.

The changes allows achieving the following usage:

```go
handler = splunkhttp.NewHandler(handler, "server", otelhttp.WithTracerProvider(provider), splunkhttp.WithTraceResponseHeader(false))
```
`splunkhttp.WithOTelOpts()` is no longer needed.

Deeper analysis and comparison to other possible approaches can be found here: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746

Moreover, some tests were moved to appropriate files (see comments in PR).


instrumentation/net/http/splunkhttp/handler_test.go
```diff
@@ -23,23 +23,22 @@ import (
 	"go.opentelemetry.io/otel/oteltest"
 )
 
-func TestTraceResponseHeaderMiddleware(t *testing.T) {
```
pellared, line 26: I moved this test to `trace_response_header_test.go`

instrumentation/net/http/splunkhttp/trace_response_header_test.go
```diff
@@ -23,12 +23,13 @@ import (
 	"go.opentelemetry.io/otel/oteltest"
 )
 
-func TestNewHandlerTraceResponseHeaderDisabled(t *testing.T) {
```
pellared, line 26: I moved this test to `handler_test.go`

owais:
image

Why is the one supposed to be more beautiful highlighted with nice colors and the old one b/w? I see what you did there. I'll remember this trick :D

owais:
Really like the net effect it has on API. Implementation looks fine and can be improve/changed later. 👍

pellared:
@owais @MrAlias 
I think we can merge it if you like this PR. All other proposals in https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746 have the same exported API. 

Moreover, we could publish the first release of https://github.com/signalfx/splunk-otel-go 😉

pellared:
@owais @MrAlias 

I feel convinced to refactor [this](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-820580882) proposal. I put rationale further down in the issue. What do you think?

owais:
That looks pretty good to me. Only concern is that this works with the HTTP handler. Will it work with other types of instrumentations as easily?

pellared:
> That looks pretty good to me. Only concern is that this works with the HTTP handler. Will it work with other types of instrumentations as easily?

This PR will not.

But [this](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/746#issuecomment-820580882) will. I will make a PR on Monday

owais:
Yeah, I meant the proposal. 👍

pellared:
@owais Here is the new PR https://github.com/signalfx/splunk-otel-go/pull/41

PR #71, opened 2021-07-22
Use goyek and go-pipeline

pellared:
## Why

Currently, we contribute to a lot of Go repositories without any unified or easy to reuse a build pipeline:

https://github.com/signalfx/?q=&type=&language=go&sort=
https://github.com/splunk/?q=&type=&language=go&sort=
https://github.com/open-telemetry/?q=&type=&language=go&sort= 
Copy-pasting fragments of Makefiles between repositories can be annoying and they are often not working on Windows.

The goal is to create a minimal common (but extensible), reusable build pipeline for Go codebases. 
The plan is to use https://github.com/goyek/goyek to create the pipeline in Go itself and use the power of Go Modules for versioning and sharing across multiple repositories.

## What

- Adoption of https://github.com/goyek/goyek and https://github.com/signalfx/go-pipeline
- Add codecov integration.

## Getting started

A few notes on how to use it in practice. After pulling the PR run `./goyek.sh -h` (or `.\goyek.ps1 -h` in PowerShell) to the CLI usage information. The build pipeline can be edited in `build/build.go` - you can for example add more tasks there.

**2.5 min demo:** https://drive.google.com/file/d/1SvUHi-rQTAzpoL-XlbFTv0WaL4ButrUt/view?usp=sharing

## Not covered

- The compatibility tests are still in Makefile as the `test` task requires Go 1.16 at the moment.
- Pre-release and some other targets (more time is needed).


Makefile
```diff
@@ -17,28 +17,20 @@ SHELL := /bin/bash
...
-# ALL_MODULES includes ./* dirs (excludes . and ./internal/tools dir).
+# ALL_MODULES includes ./* dirs (excludes . and ./build dir).
 ALL_MODULES := $(shell find . -type f -name "go.mod" -exec dirname {} \; | sort )
 # All directories with go.mod files related to opentelemetry library. Used for building, testing and linting.
-ALL_GO_MOD_DIRS := $(filter-out $(TOOLS_MODULE_DIR), $(ALL_MODULES))
+ALL_GO_MOD_DIRS := $(filter-out $(BUILD_DIR), $(ALL_MODULES))
 # All directories sub-modules. Used for tagging and generating dependabot config.
 SUBMODULES = $(filter-out ., $(ALL_GO_MOD_DIRS))
 
-.DEFAULT_GOAL := all
```
MrAlias, line 34: Why no default?

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: build
```
MrAlias, line 1: Build implies a Go binary is being built here which doesn't seem accurate.

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  ubuntu:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: actions/setup-go@v2.1.3
+        with:
+          go-version: '1.16'
+      - name: Build
+        run: ./goyek.sh -v -ci
```
MrAlias, line 18: Can this be put into a Makefile target to allow users to run this locally?

Makefile
pellared: I will add a new `goyek` target as a default.

.github/workflows/build.yml
pellared: I just decided to create a default `goyek` make target that executes `./goyek.sh`. Passing flags via Make is not fun 😉

.github/workflows/build.yml
pellared: Renamed to `goyek`

Makefile
```diff
@@ -17,48 +17,32 @@ SHELL := /bin/bash
...
-
-TOOLS = $(CURDIR)/.tools
-
-$(TOOLS):
-	@mkdir -p $@
-$(TOOLS)/%: | $(TOOLS)
-	$Q cd $(TOOLS_MODULE_DIR) \
-		&& $(GO) build -o $@ $(PACKAGE)
-
-GOLANGCI_LINT = $(TOOLS)/golangci-lint
```
MrAlias, line 58: Can we add a target to install `goyek`?

Makefile
```diff
@@ -121,42 +88,6 @@ push-tag: # example usage: make push-tag remote=origin tag=v1.100.1
 	 	git push $(remote) "$${dir:2}/$(tag)"); \
 	done
 
-DEPENDABOT_PATH=./.github/dependabot.yml
-.PHONY: gendependabot
-gendependabot: # generate dependabot.yml
```
MrAlias, line 126: Is this replaced with `goyek`? Can we keep it if not?

Makefile
```diff
@@ -121,42 +88,6 @@ push-tag: # example usage: make push-tag remote=origin tag=v1.100.1
...
-	   fi
-
-.PHONY: mod-tidy
-mod-tidy: # go mod tidy for all modules
-	${call for-all-modules,$(GO) mod tidy}
-	@echo "$(GO) mod tidy in $(TOOLS_MODULE_DIR)"
-	$Q (cd $(TOOLS_MODULE_DIR) && $(GO) mod tidy)
-
-.PHONY: diff
-diff:
```
MrAlias, line 156: Is this part of the `goyek` workflow? If not can we keep it?

Makefile
pellared: https://github.com/signalfx/go-pipeline/blob/427de9c7dae208613be08ba02995b97b68a75b71/pipeline.go#L257-L282

Makefile
pellared: I will revert it

Makefile
pellared: https://github.com/signalfx/go-pipeline/blob/427de9c7dae208613be08ba02995b97b68a75b71/pipeline.go#L257-L282

goyek.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
MrAlias, line 1: Can this live in the `build` directory?

build/build.go
```diff
@@ -12,9 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tools
```
MrAlias, line 15: Any reason to not have the build directory live in `internal`?

goyek.sh
pellared: It would be less user-friendly as `goyek.sh` and `goyek.ps1` are convenience scripts to run the build pipelines.
Then the user would have to run `./build/goyek.sh`  or `./internal/build/goyek.sh` to run the default task. I think for similar reason the `Makefile` is in the root so that the user does not have run e.g. `make -f ./build/Makefile`

build/build.go
pellared: Why I decided to NOT put it in `internal`:

1. Less nesting
2. In fact it does not expose anything as `package main` cannot be imported.
3. "Follows" https://github.com/golang-standards/project-layout 
4. [Nuke](https://nuke.build/index.html) which I discovered after creating `goyek` also contains by default creates build pipeline source code in `build` directory
5. K8s also has stuff related with building in `build` directory: https://github.com/kubernetes/kubernetes/tree/master/build

The OTel Go repos are the first place where I saw the `tools` package inside the `internal` directory.

MrAlias (approved):

pellared:
@flands Can you change the required GH status checks in `main` branch protection rules to the ones introduced in this PR?

PR #100, opened 2021-09-15
Add instrumentation for confluent-kafka-go

MrAlias:
This new module instruments the built-in`confluent-kafka-go/kafka` package. This instrumentation wraps the `Consumer` and `Producer` from the instrumented package. Kafka messages received and sent, both synchronously and asynchronously, are traced. The message headers are used to propagate trace context with a configurable `TextMapPropagtor`.

- [x] Comply with OpenTelemetry [database semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/semantic_conventions/database.md)
- [x] Package documentation


MrAlias:
@pellared thoughts on how we can exclude this package from being built on certain systems? [confluent-kafka-go is not supported on Windows](https://github.com/confluentinc/confluent-kafka-go#installing-librdkafka) and it looks like [Ubuntu 20.04 does not have a build of librdkafka-dev for i386 machine architecture](https://packages.ubuntu.com/focal/librdkafka-dev).

pellared:
> @pellared thoughts on how we can exclude this package from being built on certain systems? [confluent-kafka-go is not supported on Windows](https://github.com/confluentinc/confluent-kafka-go#installing-librdkafka) and it looks like [Ubuntu 20.04 does not have a build of librdkafka-dev for i386 machine architecture](https://packages.ubuntu.com/focal/librdkafka-dev).

For Windows I would try:

```
// +build !linux

package ...
```

For Ubuntu, I would add a step in GH workflow to install the package.

MrAlias:
> For Ubuntu, I would add a step in GH workflow to install the package in the jobs where it is needed.

Tried that. The package could not be found :|

MrAlias:
> // +build !linux

Guessing you meant `// +build !windows`. This does not work though: `level=error msg="Running error: context loading failed: no go files to analyze"`.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -0,0 +1,212 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	traceapi "go.opentelemetry.io/otel/trace"
+)
+
+var (
+	testGroupID = "gotest"
+	testTopic   = "gotest"
+)
+
+func TestConsumerChannel(t *testing.T) {
```
pellared, line 41: I suggest removing this test. It takes always 10 seconds because we cannot commit the message. Moreover, it is very artificial.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -0,0 +1,212 @@
...
+        --name kafka \
+        --network confluent \
+        -p 9092:9092 \
+        -e KAFKA_ZOOKEEPER_CONNECT=zookeeper:2181 \
+        -e KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://localhost:9092 \
+        -e KAFKA_LISTENERS=PLAINTEXT://0.0.0.0:9092 \
+        -e KAFKA_CREATE_TOPICS=gotest:1:1 \
+        -e KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1 \
+        confluentinc/cp-kafka:5.0.0
+*/
```
pellared, line 131: I prefer to use something like https://github.com/ory/dockertest

The reasons are:
- no need to create special Make targets. one can just use `go test` and it works. We could remove the `INTEGRATION` env var.
- does not require Docker CLI so often you can run such tests inside a `golang` docker container if you just mount `-v /var/run/docker.sock:/var/run/docker.sock` (here I am not sure if it will work b/c of native dependencies)

pellared (commented):
I only made a quick review so you can get any feedback.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
pellared: I just learned that this approach is even recommended by Docker 😉 
See: https://docs.docker.com/language/golang/run-tests/

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/config.go
```diff
@@ -0,0 +1,84 @@
...
+// this instrumentation library.
+func WithTracerProvider(tp trace.TracerProvider) Option {
+	return optionFunc(func(c *config) {
+		c.Tracer = tp.Tracer(
+			instrumentationName,
+			trace.WithInstrumentationVersion(splunkotel.Version()),
+		)
+	})
+}
+
```
pellared, line 76: Maybe worth adding `WithAttributes`?

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+func TestConsumerPoll(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+	if _, ok := os.LookupEnv("INTEGRATION"); !ok {
+		t.Skip("to enable integration test, set the INTEGRATION environment variable")
+	}
+
```
pellared, line 140: What do you think of adding `defer goleak.VerifyNone(t)` from https://github.com/uber-go/goleak to have an additional safety net against concurrency bugs?

PS. I reviewed the concurrency design in production code and looks good 👍

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
```diff
@@ -0,0 +1,148 @@
...
+	return wrapped
+}
+
+func (p *Producer) traceProduceChannel(out chan *kafka.Message) chan *kafka.Message {
+	if out == nil {
+		return nil
+	}
+
+	in := make(chan *kafka.Message, cap(out))
+	go func() {
```
pellared, line 65: EOL of this goroutine is not synchronized with `Close` so users can miss the last `span.End()`.

I think something like this is safer (better synchronized):

```go
// in func (p *Producer) traceProduceChannel:
p.wg.Add(1) // wg is a new field in Producer to synchronize EOL of the goroutine scheduled below
go func() {
  for { 
    select {
      case <-p.closeSig:
        p.wg.Done()
        return
      case msg := <- in
        span := p.startSpan(msg)
        out <- msg
        span.End()
}()

// Close calls the wrapped Producer.Close and closes the producer channel.
func (p *Producer) Close() {
	close(p.produceChannel)
	p.Producer.Close()
	p.wg.Wait()
}
```

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
```diff
@@ -0,0 +1,148 @@
...
+// Produce calls the wrapped Producer.Produce and traces the request.
+func (p *Producer) Produce(msg *kafka.Message, deliveryChan chan kafka.Event) error {
+	span := p.startSpan(msg)
+
+	// if the user has selected a delivery channel, we will wrap it and
+	// wait for the delivery event to finish the span
+	if deliveryChan != nil {
+		oldDeliveryChan := deliveryChan
+		deliveryChan = make(chan kafka.Event)
+		go func() {
```
pellared, line 116: should we not have here:

```go
		p.wg.Add(1)
		go func() {
			defer p.wg.Done()
```

?

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
```diff
@@ -0,0 +1,148 @@
...
+				span.RecordError(err)
+				span.SetStatus(codes.Error, err.Error())
+			}
+			span.End()
+			oldDeliveryChan <- evt
+		}()
+	}
+
+	err := p.Producer.Produce(msg, deliveryChan)
+	// with no delivery channel, finish immediately
```
pellared, line 133: shouldn't we wrap `Events()` to handle this scenario? I think it is a quite regular use case just to do "fire-and-foget"

See: https://pkg.go.dev/github.com/confluentinc/confluent-kafka-go@v1.7.0/kafka?utm_source=gopls#Producer.Produce

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Close calls the wrapped Producer.Close and closes the producer channel.
+func (p *Producer) Close() {
+	close(p.produceChannel)
+	p.Producer.Close()
+}
+
+// Produce calls the wrapped Producer.Produce and traces the request.
+func (p *Producer) Produce(msg *kafka.Message, deliveryChan chan kafka.Event) error {
```
pellared, line 108: I think this instrumentation does not handle `"go.delivery.reports": false,` scenario. If so we should at least mention it in the README.md.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
pellared: moreover, thanks to this change the caller goroutine should panic if it tries ti send a message after closing the producer

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/consumer.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (c *Consumer) traceEventsChannel(in chan kafka.Event) chan kafka.Event {
+	// in will be nil when consuming via the events channel is not enabled.
+	if in == nil {
+		return nil
+	}
+
+	out := make(chan kafka.Event, cap(in))
+	go func() {
```
pellared, line 79: should we not have here:

```go
		p.wg.Add(1)
		go func() {
			defer p.wg.Done()
```

and add `p.wg.Done()` in `Close` method?

pellared (dismissed):
I think we can merge it and address the comments in another PR(s).

.github/dependabot.yml
```diff
@@ -11,7 +11,11 @@ updates:
     schedule:
       interval: "daily"
   - package-ecosystem: "gomod"
-    directory: "/build"
```
pellared, line 14: why? what if there is a new release of `goyek`?

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
pellared: I think it is not worth addressing it in this PR.
I created an issue out of it: https://github.com/signalfx/splunk-otel-go/issues/101
I would be thankful if you express your opinion there 😉

Feel free to resolve this comment.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
pellared: You can create an issue out of it if you do not want to spend time on this right now.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
pellared: You can create an issue out of it if you do not want to spend time on this right now.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
pellared: You can create an issue out of it if you do not want to spend time on this right now.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/consumer.go
pellared: You can create an issue out of it if you do not want to spend time on this right now.

pellared (approved):
Approving as I do not want to block this PR.

.github/dependabot.yml
MrAlias: Fixed with https://github.com/signalfx/splunk-otel-go/pull/117

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -47,7 +46,54 @@ var (
...
+	partition := int32(0)
+	sr, opts := newFixtures()
+	p := newProducer(t, opts...)
+
+	done := make(chan struct{})
+	var sent *kafka.Message
+	go func() {
+		defer close(done)
+		sent = requireEventIsMessage(t, <-p.Events())
```
pellared, line 60: we should not use `require` or `t.Fatalf/FailNow` in new goroutines.
See: https://pkg.go.dev/testing#T.FailNow

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.

I prefer having all assertions in the main testing goroutine if it does not complicate the code too much. But usually it s nicer to have all assertions grouped.

When looking at this code I am not even sure if spawning a new goroutine to receive the message is needed at all

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -47,7 +46,54 @@ var (
...
+			Key:   key,
+			Value: val,
+		}
+	}()
+
+	// Wait for the delivery report goroutine to finish.
+	<-done
+	require.NoError(t, sent.TopicPartition.Error)
+
+	p.Close()
```
pellared, line 77: is it imortant that this `p.Close()` is here?
maybe we could `defer p.Close()` after `p := newProducer(t, opts...)`?

same for `TestFunctionBasedProducer`

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func requireEventIsMessage(t *testing.T, e kafka.Event) *kafka.Message {
+	m, ok := e.(*kafka.Message)
+	require.Truef(t, ok, "invalid response from Kafka: %v", e)
+	require.NoError(t, m.TopicPartition.Error)
+	return m
+}
+
+func TestMain(m *testing.M) {
```
pellared, line 212: nit: can you move it to the top (as the first function)?

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
```diff
@@ -0,0 +1,325 @@
...
+				Topic:     &testTopic,
+				Partition: partition,
+			},
+			Key:   key,
+			Value: val,
+		}
+	}()
+
+	// Wait for the delivery report goroutine to finish.
+	<-done
```
pellared, line 74: sometime a create a "test timeout channel" to prevent the test from deadlocking

we can do it e.g. like this

```go
// ...
  timeout := timeoutCh(t, time.Minute)
// ...
	select {
	case <-timeout:
		t.Fatal("test timeout")
	case <- input:
		// we can put the assertions here or just assign the value to a variable to decrease the indentation
	}
// ...

// timeoutCh prepares a channel that can be used to implement a test timeout.
func timeoutCh(t *testing.T, d time.Duration) <-chan time.Time {
	timeout := time.NewTimer(d)
	t.Cleanup(func() { timeout.Stop() })
	return timeout.C
}
```

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
MrAlias: I'm not sure it is possible or maybe just not worth it. If there is no delivery channel we need to track the state of the enqueued message. Otherwise, we will not be able to ensure the span ended corresponds to the message we just enqueued. Given we do not know the offset of the message when we enqueue the message here, I'm not sure we could ever actually re-associate the span we create here with the response message.

It might be possible because the underlying implementation looks to return the same pointer to the enqueued message, but there is no guarantee this implementation will continue.

I think the best we can do is what is done here - end the span once we have successfully enqueued it.

Feel free to open an issue to track this if you think there is indeed a way to handle this.

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/test/kafka_test.go
MrAlias: `Close` flushes any state held by the underlying `Producer`/`Consumer`. Having it called before we evaluate any spans ensures that all spans are going to be accounted for, both now and if any others are added in the future.

It seems like a good idea to keep the call where it is for these reasons, no?

pellared: You are totally right 👍 Forgot about that

instrumentation/github.com/confluentinc/confluent-kafka-go/kafka/splunkkafka/producer.go
MrAlias: Looking a the existing solution, I think we suffer from the same issue when the user provides a delivery channel. There is no guarantee they do not provide that channel to multiple calls to `Produce`. Meaning that our assumed guarantee that the response returned on that deliver channel is going to be associated with that specific call is incorrect at times.

PR #124, opened 2021-09-24
Document the splunksql package

### LINKED ISSUE #103: Add README to splunksql package
_Originally posted by @pellared in https://github.com/signalfx/splunk-otel-go/issues/88#issuecomment-917890784_

MrAlias:
Resolve #103


instrumentation/database/sql/splunksql/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This package is designed to be used in conjunction with existing database
+drivers that have been instrumented so OpenTelemetry semantic conventions can
+be discovered. An example of this type of use can be found
+[here](../../../github.com/go-sql-driver/mysql/splunkmysql).
+
+This package can be used directly as well. It becomes the users responsibility
+to ensure accurate and complete information about the database system is passed
+as attributes to ensure OpenTelemetry semantic conventions are satisfied. An
+example of this can be found [here](./example_test.go).
```
pellared, line 18: what about `./example_sql_test.go`? I think it would be good to mention it or combine those two examples.

pellared (approved):
BTW WDYT about https://github.com/signalfx/splunk-otel-go/pull/88#issuecomment-920790593?

pellared:
Do you think it would be valuable to documented what spans, attributes, events are created using this package?

MrAlias:
> Do you think it would be valuable to documented what spans, attributes, events are created using this package?

This seems like something that would become stale easily. Having stale docs can cause more harm then good.

MrAlias:
> BTW WDYT about [#88 (comment)](https://github.com/signalfx/splunk-otel-go/pull/88#issuecomment-920790593)?

I like the idea, but I'm not entirely sure how to include an attribute from the context without adding some additional functionality. I'd be interested in seeing an example if you have one, but I would also not want to block this PR on including that. What do you think @pellared?

instrumentation/database/sql/splunksql/README.md
MrAlias: The `example_sql_test.go` file contains info on how to register an `InstrumentationConfig`. It is rendered in the go docs:



Having it structured this way that example is put into the appropriate context for the user looking for information about how to use the `Register` function. In the paragraph above I link to the complete example from the `splunksql` package of how to setup instrumentation this way. I can add something in that paragraph about the other example file, but I'm not sure it is going to be the best in the long term. Having the Go docs handle this seems sufficient. What do you think @pellared?

MrAlias: We can also link to the pkg.go.dev docs from the README. What do you think of that?

pellared: I think what is now should be good enough. If someone will ask us for help in future it would mean that the docs should be improved.

PR #155, opened 2021-10-13
Add miekg/dns instrumentation

MrAlias:
This new module instruments the `github.com/miekg/dns` package.

- [x] Comply with OpenTelemetry semantic conventions (there are currently no requirements)
- [x] Package documentation
- [x] CHANGELOG.md updated


instrumentation/github.com/miekg/dns/splunkdns/test/client_test.go
```diff
@@ -0,0 +1,168 @@
...
+		Net:        pc.LocalAddr().Network(),
+		PacketConn: pc,
+		Handler:    mux,
+	}
+	go func() {
+		_ = server.ActivateAndServe()
+	}()
+	require.NoError(t, serverUp(pc.LocalAddr().String(), time.Second*10))
+
+	return server
```
pellared, line 61: probably you could use `t.Cleanup` so that the caller would not need to remember to shutdown the server

instrumentation/github.com/miekg/dns/splunkdns/test/client_test.go
```diff
@@ -0,0 +1,168 @@
...
+	})
+
+	server := &dns.Server{
+		Addr:       pc.LocalAddr().String(),
+		Net:        pc.LocalAddr().Network(),
+		PacketConn: pc,
+		Handler:    mux,
+	}
+	go func() {
+		_ = server.ActivateAndServe()
```
pellared, line 57: you could also create a `chan error` and get the error after calling shutdown
then you could double-check the result and make sure that this goroutine is closed (and it could be done in t.Cleanup probably)

pellared (commented):
Entry in `CHANGELOG.md` is missing 😉 

I have reviewed it quickly. It would be better if @TomRoSystems would give it another pass.

pellared (approved):

instrumentation/github.com/miekg/dns/splunkdns/test/client_test.go
```diff
@@ -0,0 +1,163 @@
...
+		Net:        pc.LocalAddr().Network(),
+		PacketConn: pc,
+		Handler:    mux,
+	}
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- server.ActivateAndServe()
+	}()
+	t.Cleanup(func() {
+		assert.NoError(t, <-errCh, "failed to start server")
```
pellared, line 61: nit

```suggestion
		assert.NoError(t, <-errCh, "server.ActivateAndServe returned an error")
```

as there may be multiple reasons why it returns an error

moreover I hope this error message would make it easier to find where the error come from

PR #226, opened 2021-12-02
Unify the Config type across instrumentation

MrAlias:
The Config and its related Options are recreated for all our instrumentation currently. This results in a lot of duplicate code and tests. This adds a common Config type with appropriate functionality as well as common options. These internal types and functions are expected to be reused in instrumentation libraries by embedding the types or forwarding calls to functions.


pellared (commented):
Can you please update at least one instrumentation library to see how the new internal package would be used?

MrAlias:
> Can you please update at least one instrumentation library to see how the new internal package would be used?

Added a use-case in the `splunkleveldb` instrumentation. PTAL

instrumentation/github.com/syndtr/goleveldb/leveldb/splunkleveldb/iterator.go
```diff
@@ -32,10 +32,10 @@ type iter struct {
 // called.
 func WrapIterator(it iterator.Iterator, opts ...Option) iterator.Iterator {
 	c := newConfig(opts...)
-	sso := c.mergedSpanStartOptions(
+	sso := c.MergedSpanStartOptions(
 		trace.WithAttributes(semconv.DBOperationKey.String("Iterator")),
 	)
-	_, span := c.resolveTracer().Start(c.ctx, "Iterator", sso...)
+	_, span := c.ResolveTracer(c.ctx).Start(c.ctx, "Iterator", sso...)
```
pellared, line 38: I find it awkward to have the the `ctx` in the config struct. 
I propose to have two methods in the common config:
- ResolveTracer
- ResolveTracerContext (when some call have context)

Same for `WithSpan`. 

Then if the instrumented library used context we use `ResolveTracerContext` and `WithSpanContext` (e.g. for `http`) if not then we simply use `ResolveTracerContext` and `WithSpan` (e.g. for `leveldb`).

Calling `trace.SpanFromContext(ctx)` is useless if we know that it will never return anything.

If you disagree I would rather put `context.Background()` in all places instead of `cfg.ctx`. Current code can give a false impression that can be some real context.

pellared (approved):
Even though I dislike type embedding it looks like the most efficient way to avoid code duplication in this scenario. Left a few comments but in general, I think that this is a very pragmatic approach 👍

instrumentation/github.com/syndtr/goleveldb/leveldb/splunkleveldb/iterator.go
MrAlias: > I find it awkward to have the the `ctx` in the config struct.

Sure, agreed. However, this is an artifact of the underlying library not accepting a context as an argument. The instrumentation then needs to accept this an apply it as needed.

> I propose to have two methods in the common config:
> 
> * ResolveTracer
> * ResolveTracerContext (when some call have context)

`ResolveTracer` is equivalent to just `Config.Tracer`. If you don't have a context to extract a `TracerProvider` from it should not be used.

> Same for `WithSpan`.

The context passed to `WithSpan` is updated to include the span and passed to the closure. Having the ability to set the parent is going to be important for more than just span propagation. Like the rest of the Go ecosystem, having a context as a first argument is a good idea in that it keeps context propagation explicit and adding `*Context` type functions are useful only in that they allow backwards compatibility for context propagation. Having one from the start seems to be the opposite of idiomatic Go.

> If you disagree I would rather put `context.Background()` in all places instead of `cfg.ctx`. Current code can give a false impression that can be some real context.

That can in fact be a real context, with a span embedded in it. Passing that context as included in this PR ensures the correct user specified `TracerProvider` is used an heredity is preserved.

instrumentation/github.com/syndtr/goleveldb/leveldb/splunkleveldb/config.go
```diff
@@ -17,35 +17,38 @@ package splunkleveldb
...
 
 // instrumentationName is the instrumentation library identifier for a Tracer.
 const instrumentationName = "github.com/signalfx/splunk-otel-go/instrumentation/github.com/syndtr/goleveldb/leveldb/splunkleveldb"
 
 // config contains tracing configuration options.
 type config struct {
-	ctx              context.Context
-	tracer           trace.Tracer
-	defaultStartOpts []trace.SpanStartOption
+	*internal.Config
```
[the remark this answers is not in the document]
MrAlias: The returned value from `NewConfig` is a pointer.

If we want to optimize I would want to build out benchmarks first. Definitely something we can do in the future if needed.

PR #295, opened 2021-12-17
Support OTEL_PROPAGATORS in distro

MrAlias:


MrAlias:
Going to update docs and change to a lazy evaluation of the default props.

distro/otel.go
```diff
@@ -69,10 +73,16 @@ func Run(opts ...Option) (SDK, error) {
...
-	global.SetTracerProvider(traceProvider)
+	otel.SetTracerProvider(traceProvider)
 
-	// TODO: add and honor option to set additional propagators.
-	global.SetTextMapPropagator(b3.New())
+	if c.Propagator != nil {
+		if c.Propagator == nonePropagator {
+			// Set to an empty propagator if none was specified
+			otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator())
+		} else {
```
pellared, line 82: how about implementing `nonePropagatorType` as a NullObject (noop implementation) so that this `if` statement would not be required?

CHANGELOG.md
```diff
@@ -59,6 +59,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add the
   `github.com/signalfx/splunk-otel-go/instrumentation/github.com/gomodule/redigo/splunkredigo`
   instrumentation for the `github.com/gomodule/redigo` package. (#288)
+- Add the `WithPropagator` option to
+  `github.com/signalfx/splunk-otel-go/distro` along with parsing of the
+  OTEL_PROPAGATORS environment variable to set the global OpenTelemetry
```
pellared, line 64:
```suggestion
  `OTEL_PROPAGATORS` environment variable to set the global OpenTelemetry
```

pellared (commented):
I made a quick look. Looks very nice 👍 
@TomRoSystems PTAL

distro/otel.go
MrAlias: By making `nonePropagatorType` a no-op implementation and then setting the global propagator to that value it would contradict the behavior of what setting `OTEL_PROPAGATORS` to `none` is intended to accomplish (i.e. no automatically configured propagator).

theletterf (approved):

distro/otel.go
pellared: Would it change anything for the user? What is the difference between an empty composite propagator or a composite propagator with a noop propagator.

pellared: Moreover take notice that even in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#enum-value the `none` is described as "a null object (empty, no-op) value"

MrAlias: > Moreover take notice that even in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#enum-value the `none` is described as "a null object (empty, no-op) value"

Sure, but please take note of [the specific specification for `OTEL_PROPAGATOR=none`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration):

> `none`: No automatically configured propagator.

Hence the omission of automatically configuring a propagator.

pellared: I see. Feel free to resolove is setting  OTEL_PROPAGATORS=tracecontext,none **does not** result in a crash at runtime

README.md
```diff
@@ -33,7 +33,7 @@ Table of Contents:
 
 This Splunk distribution comes with the following defaults:
 
-- [B3 context propagation](https://github.com/openzipkin/b3-propagation).
+- [W3C tracecontext propagation](https://www.w3.org/TR/trace-context/).
```
pellared, line 36: and W3C baggage https://www.w3.org/TR/baggage/

distro/README.md
```diff
@@ -0,0 +1,72 @@
...
+- Default value: empty (i.e. `""`)
+- Environment variable: `SPLUNK_ACCESS_TOKEN`
+
+### `WithEndpoint`
+
+`WithEndpoint` configures the Splunk endpoint that telemetry is sent to.
+
+- Default value: depends on the exporter used.
+  - For the `otlp` over gRPC exporter: `"localhost:4317"`
+  - For the `jaeger-thrift-splunk` exporter: `"http://127.0.0.1:9080/v1/trace"`
```
pellared, line 47: how about documenting the env vars that can be used to set these values?

distro/config.go
```diff
@@ -62,6 +86,81 @@ func (c config) Validate() error {
...
+	},
+	// OpenTracing Trace
+	"ottrace": func() propagation.TextMapPropagator {
+		return ot.OT{}
+	},
+	// None, explicitly do not set a global propagator.
+	"none": func() propagation.TextMapPropagator {
+		return nonePropagator
+	},
+}
```
pellared, line 130: how about simplifying it to a function like

```go
func createPropagator(name string) propagation.TextMapPropagator {
  switch name {
  case "tracecontext":
    return propagation.TraceContext{}
  }
// etc
}
```

distro/config.go
```diff
@@ -62,6 +86,81 @@ func (c config) Validate() error {
...
+
+func loadPropagator(name string) propagation.TextMapPropagator {
+	var props []propagation.TextMapPropagator
+	for _, part := range strings.Split(name, ",") {
+		factory, ok := propagators[part]
+		if !ok {
+			// Skip invalid data.
+			continue
```
pellared, line 138: can we return an error? 

If something goes wrong during initialization then we are returning an error: https://github.com/signalfx/splunk-otel-go/blob/main/distro/otel.go

distro/README.md
MrAlias: Yeah, it looks like I included it in the follow on PR (https://github.com/signalfx/splunk-otel-go/pull/300/files#diff-f5ff2011d061002f61a86d92d59293383700444948d773a18a76625dd5986030). I'll move those changes forward.

distro/config.go
MrAlias: I mean, sure. There are probably 20 other ways we could also do it. However, the current approach isolates the mapping of environment variable value to TextMapPropagator from the parsing of the combined environment variable values. Having this separation makes the code easier to read.

distro/config.go
MrAlias: What would we do with the error? Do we want to fail if this situation is encountered, or ultimately just move on?

pellared: Personally I would suggest failing as this would be an unsupported configuration

distro/config.go
pellared: This is very minor. I will make a PR when I have time.

distro/config.go
```diff
@@ -62,6 +86,81 @@ func (c config) Validate() error {
 	return nil
 }
 
+type nonePropagatorType struct{ propagation.TextMapPropagator }
+
+// nonePropagator signals the disablement of setting a global
+// TextMapPropagator.
+var nonePropagator = nonePropagatorType{}
```
pellared, line 93: nit: maybe just use `nil` instead? having `nonePropagatorType` gives a false impression that the interface is implemented. It would solve "my problems" that `nonePropagator` is not a "NullObject" having a no-op implementation. `nil` is nothing - very obvious 😄 feel free to resolve if you disagree or do not want to lose time on it

MrAlias: Yeah, that's a good idea, and one I looked into. The issue that it runs into is that when the user passes `nil` to the `WithPropagator` meaning setting a propagator should be disabled and at the same time `OTEL_PROPAGATOR` is defined as non-`"none"`. Without a sentinel value to signal this disablement from the first option it is indistinguishable from not setting that option.

pellared: I see. Thanks. Can you add this info as a comment? 😉

pellared: Probably as an alternative you can filter out the `nil` before passing to `WithPropagator`.

MrAlias: @pellared I'm definitely open to suggestions if you have them on how to resolve this issue and still use the `nil` value. I agree it would clean things up nicely.

MrAlias: > Probably as an alternative you can filter out the `nil` before passing to `WithPropagator`.

I'm not sure I follow. Do you mean we could change the behavior and drop any passed nil?

> Can you add this info as a comment?

I added [this](https://github.com/signalfx/splunk-otel-go/pull/295/files#diff-8d2c05f32348fdabc91f4819ba3cf985b618d684a8c79c1111b7c8794164be92R220-R222). Did you have a suggestion for more?

distro/config.go
MrAlias: We need to add logging support in the future to be compliant with other areas of the GDI specification. This seems likes it might warrant a log message. What do you think? I can add a comment to fix in the future if you think so.

distro/config.go
pellared: That is fine. Thanks a lot 👍

distro/config.go
pellared: The error-handling my preference is:
- fail-fast and return error when the issue is connected with bad initialization
- log when the issue is "at runtime" - however, these issues are often transient and can make "a mess" in logs so this can be revisited later (I remember that the OTel collector has some good guidelines here)

It is compliant with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles

> The API or SDK may fail fast and cause the application to fail on initialization, e.g. because of a bad user config or environment, but MUST NOT cause the application to fail later at run time, e.g. due to dynamic config settings received from the Collector.

Some functions in OTel SDK are already returning errors when the provided data is wrong e.g. [here](https://github.com/signalfx/splunk-otel-go/blob/9a017c8ef6c1553cce1a6dd7abfd5e093b5df6e0/distro/otel.go#L62)

Whatever approach/guidelines regarding error-handling you (or we) prefer I suggest documenting it in `CONTRIBUTING.md` (of course in a separate PR). Here is a simple example we have in OTel .NET Auto-Instrumentation: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/DESIGN.md#error-handling It will facilitate future discussions, reviews, development etc

pellared: I am OK leaving a TODO/FIXME comment if you want to think about it more or address it later.

README.md
```diff
@@ -33,7 +33,8 @@ Table of Contents:
 
 This Splunk distribution comes with the following defaults:
 
-- [B3 context propagation](https://github.com/openzipkin/b3-propagation).
+- [W3C tracecontext](https://www.w3.org/TR/trace-context/) and
+  [baggage](https://www.w3.org/TR/baggage/) context propagation.
```
pellared, line 37: nit:

```suggestion
  [W3C baggage](https://www.w3.org/TR/baggage/) context propagation.
```

distro/README.md
```diff
@@ -0,0 +1,79 @@
+# Package `github.com/signalfx/splunk-otel-go/distro`
+
+This package provides a Splunk distribution of the OpenTelemetry Go SDK. It is
+designed to provide an SDK properly configured to be used with the Splunk
+platform out-of-the-box.
+
+## Getting Started
+
+The main entry point for the package is the [`Run`][] function. Use this
```
pellared, line 9: nit: I think you can change
```
[`Run`][]
```
to 
```
[`Run`]
```
but you would have to test it 😄 

this is at least what I was always familiar with and is used here: https://raw.githubusercontent.com/signalfx/splunk-otel-go/main/CHANGELOG.md

distro/README.md
```diff
@@ -0,0 +1,79 @@
...
+function to create an SDK that is ready to be used with OpenTelemetry and
+forward all telemetry to Splunk. See [`example_test.go`](./example_test.go) for
+a complete example.
+
+## Configuration
+
+The [`SDK`][] is configured with the following options.
```
pellared, line 16: same as above

distro/README.md
```diff
@@ -0,0 +1,79 @@
...
+Environment variable can be used to set related option values, but the value
+set in code will take precedence. This is the same behavior the default
+OpenTelemetry SDK has.
+
+The following sections contain specific information for each option.
+
+### `WithAccessToken`
+
+`WithAccessToken` configures the authentication token used to authenticate
+telemetry delivery requests to a Splunk back-end.
```
pellared, line 38:
```suggestion
telemetry sent directly to Splunk Observability Cloud.
```

distro/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+  The environment variable values are restricted to the following.
+  - `"tracecontext"`: W3C tracecontext
+  - `"baggage"`: W3C baggage
+  - `"b3"`: B3 single-header format
+  - `"b3multi"`: B3 multi-header format
+  - `"jaeger"`: Jaeger
+  - `"xray"`: AWS X-Ray
+  - `"ottrace"`: OpenTracing
+  - `"none"`: None, explicitly do not set a global propagator
```
pellared, line 73:
```suggestion
  - `"none"`: None; explicitly do not set any global propagator
```

I think it will give more stress more that none propagator will be set (even if there will be more values than `none`).

side comment: OTel Java SDK considers something like `tracecontext,baggage,none` as an invalid value.

distro/config.go
```diff
@@ -84,19 +183,45 @@ func (fn optionFunc) apply(c *config) {
...
 	return optionFunc(func(c *config) {
 		c.Endpoint = endpoint
 	})
 }
 
-// WithAccessToken configures the authentication token
-// allowing exporters to send data directly to a Splunk back-end.
-// Setting empty string results in no operation.
+// WithAccessToken configures the authentication token used to authenticate
+// telemetry delivery requests to a Splunk back-end. Passing an empty string
```
pellared, line 195: I find the term `directly to Splunk Observability Cloud` more accurate and it is used in the "official docs" e.g. https://docs.splunk.com/Observability/gdi/get-data-in/application/java/instrumentation/instrument-java-application.html#send-data-directly-to-observability-cloud

pellared (approved):
nice work. just nit comments 😄

PR #311, opened 2022-01-04
Add gopkg.in/olivere/elastic instrumentation

MrAlias:
This new module instruments the `gopkg.in/olivere/elastic` module.

- [x] Comply with OpenTelemetry semantic conventions
  - This defaults to HTTP semantics (span naming, attributes) because that information will always be known, but it attempts to parse information and comply with database semantics if possible.
- [x] Package documentation
- [x] CHANGELOG.md updated


instrumentation/gopkg.in/olivere/elastic/splunkelastic/README.md
```diff
@@ -0,0 +1,11 @@
+# Splunk instrumentation for `gopkg.in/olivere/elastic`
+
+This package provides OpenTelemetry instrumentation for the
+[gopkg.in/olivere/elastic.v{3,7}](https://gopkg.in/olivere/elastic.v7)
```
pellared, line 4: Shouldn't we also instrument `github.com/olivere/elastic` and `github.com/olivere/elastic/v7` packages as these are the "supported" Go modules defined by: https://github.com/olivere/elastic#releases

instrumentation/gopkg.in/olivere/elastic/splunkelastic/elastic.go
```diff
@@ -0,0 +1,157 @@
...
+	// <db.operation>
+	return op
+}
+
+type wrappedBody struct {
+	ctx  context.Context
+	span trace.Span
+	body io.ReadCloser
+}
+
```
owais, line 136: Interesting technique. Does this mean the span will never end if the full body is not read or reader is not closed? I imagine any correctly written code would always call Close but it is rare for an instrumentation to rely on the end user writing correct code to function properly.

instrumentation/gopkg.in/olivere/elastic/splunkelastic/elastic.go
```diff
@@ -0,0 +1,157 @@
...
+var _ io.ReadCloser = (*wrappedBody)(nil)
+
+func (wb *wrappedBody) Read(b []byte) (int, error) {
+	n, err := wb.body.Read(b)
+	switch err {
+	case nil:
+		// nothing to do here but fall through to the return
+	case io.EOF:
+		wb.span.End()
```
owais, line 145: Can this happen and Close() be called on the same body object? would span handle double End() calls gracefully?

instrumentation/gopkg.in/olivere/elastic/splunkelastic/README.md
pellared: I think mentioning it explicitly here is good enough.

instrumentation/gopkg.in/olivere/elastic/splunkelastic/elastic.go
pellared: The same problem is with `http` package instrumentation. Probably we could use [`runtime.SetFinalizer`](https://pkg.go.dev/runtime#SetFinalizer) to call Close, but the overhead may be big... Maybe it could be an opt-out functionality?

pellared: More: https://medium.com/a-journey-with-go/go-finalizers-786df8e17687 
I just learned that they are used in `http` package anyway so probably this is how we should handle the scenario if the instrument code does not close the Body

pellared: However, it won't help a lot as the span times would be VERY inaccurate...

instrumentation/gopkg.in/olivere/elastic/splunkelastic/elastic.go
MrAlias: It will be implementation dependent, but the default SDK this distro uses does handle multiple calls to `End` gracefully (in that the subsequent calls are ignored).

instrumentation/gopkg.in/olivere/elastic/splunkelastic/elastic.go
MrAlias: > Interesting technique. Does this mean the span will never end if the full body is not read or reader is not closed? I imagine any correctly written code would always call Close but it is rare for an instrumentation to rely on the end user writing correct code to function properly.

Yeah, as pointed out I had copied this from the upstream HTTP instrumentation (and from our local HTTP instrumentation). But I think this comes down to how we want to model the telemetry. For HTTP libraries including the response parsing in the span timing seems appropriate, especially if you want body parsing errors to fail the span. With Elasticsearch database telemetry this might not apply. Modeling the span around the connection and remote procedure itself seems to fit better with our other database instrumentation.

I'll look into updating this to match the database model. Sound good?

owais: Sounds good!

owais (approved):

PR #332, opened 2022-01-27
Set splunk.distro.version attr in resource

MrAlias:
As per the GDI specification: https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md#splunk-resource-attributes


pellared (approved):
Consider adding a line in the CHANGELOG

PR #336, opened 2022-02-08
Add logging to the distro

MrAlias:
This adds zap logging to the `SDK` by way of a `logr` wrapper. The choice to use zap is one motivated to match what the collector is doing and by the fact that it supports rate limiting (something we may want to tune in the future).

The default logger level is configured via the `OTEL_LOG_LEVEL` envar. The default logging level is `info`.

This only adds logging to the `distro`. It does not add logging to instrumentation so as to not couple the instrumentation to the Splunk `distro` of OTel.

Tests have been updated to only output log messages for verbose test runs.

This includes the addition of logging a warning when the service name is not set, conforming with [the GDI specification](https://github.com/signalfx/gdi-specification/blob/v1.2.0/specification/configuration.md#instrumentation-libraries).


distro/README.md
```diff
@@ -11,6 +11,20 @@ function to create an SDK that is ready to be used with OpenTelemetry and
...
 a complete example.
 
+## Logging
+
+By default the [`SDK`] will log relevant operation information, warnings, and
+errors. This logging can be explicitly configured in the code that creates the
+[`SDK`] using the `WithLogger` option to configure a logger that only logs what
+is wanted. Or, the default logging level can be set using the OpenTelemetry
+defined `OTEL_LOG_LEVEL` environment variable. The following values for that
+variable are understood.
```
theletterf, line 21:
```suggestion
By default, the [`SDK`] logs relevant information, warnings, and errors. 
You can set the default logging level by setting the `OTEL_LOG_LEVEL` 
environment variable to one of the following values:
```

distro/README.md
```diff
@@ -11,6 +11,20 @@ function to create an SDK that is ready to be used with OpenTelemetry and
...
+
+- `"error"`: only log errors
+- `"warn"`: log errors and warnings
+- `"info"`: log operation information, warnings, and errors
+- `"debug"`: log debugging and operation information, warnings, and errors
+
```
theletterf, line 27:
```suggestion
- `"error"`: Log errors
- `"warn"`: Log errors and warnings
- `"info"`: Log information, warnings, and errors
- `"debug"`: Log debugging and operation information, warnings, and errors

Logging can be explicitly configured in the code that creates the [`SDK`] 
using the `WithLogger` option.
```

distro/README.md
```diff
@@ -11,6 +11,20 @@ function to create an SDK that is ready to be used with OpenTelemetry and
...
 ## Configuration
 
 The [`SDK`] is configured with the following options.
```
theletterf, line 30: We usually don't mention the SDK in other distros's repos. I assume this just describes our distribution defaults?

distro/README.md
```diff
@@ -100,5 +115,23 @@ a batch span processor.
 - Default value: none; an non-TLS connection is used.
 - Environment variable: n/a (only configurable in code).
 
+### `WithLogger`
+
+`WithLogger` configures the logger used by this distro.
+
+The [`logr.Logger`] provided should be configured with a verbosity enabled to
+emit [`Info`] logs of the desired level. The following log level to verbosity
+value are used.
```
theletterf, line 124:
```suggestion
Configure [`logr.Logger`] with a verbosity value that allows to emit
 [`Info`] logs of the desired level. The SDK uses the following log level 
 to verbosity value correspondences:
```

distro/README.md
```diff
@@ -100,5 +115,23 @@ a batch span processor.
...
+
+- warning: `0`
+- info: `1`
+- debug: `2+`
```
theletterf, line 128:
```suggestion
| Log level | Verbosity value |
|---|---|
| warning | `0` |
| info | `1` |
| debug | `2` or higher |
```

distro/README.md
```diff
@@ -100,5 +115,23 @@ a batch span processor.
...
+
+- Default value: [`zapr`]; configured for info logging.
+- Environment variable: n/a (only configurable in code).
```
theletterf, line 131:
```suggestion
The default value, [`zapr`], is configured for `info` logging. This setting is configurable only in code.
```

distro/otel.go
```diff
@@ -79,6 +88,12 @@ func Run(opts ...Option) (SDK, error) {
 	if err != nil {
 		return SDK{}, err
 	}
+	if !serviceNameDefined(res) {
+		c.Logger.Info(
+			"service.name attribute is not set, your service is unnamed and will be difficult to identify. " +
+				`Set your service name using the OTEL_SERVICE_NAME environment variable (i.e. OTEL_SERVICE_NAME="<YOUR_SERVICE_NAME_HERE>")`,
```
theletterf, line 94:
```suggestion
			"service.name attribute is not set. Your service is unnamed and might be difficult to identify. " +
				`Set your service name using the OTEL_SERVICE_NAME environment variable. For example, OTEL_SERVICE_NAME="<YOUR_SERVICE_NAME_HERE>")`,
```

distro/otel.go
```diff
@@ -58,11 +60,18 @@ func (s SDK) Shutdown(ctx context.Context) error {
...
+		c.Logger.Error(e, "OpenTelemetry error")
+	}))
+	otel.SetLogger(c.Logger)
+
 	if c.Propagator != nil && c.Propagator != nonePropagator {
 		otel.SetTextMapPropagator(c.Propagator)
 	}
 
 	if c.TraceExporterFunc == nil {
+		c.Logger.V(1).Info("tracing disabled: no trace exporter")
```
theletterf, line 74: Can we provide a suggestion to users here, similar to what you've added in 93 to 94?

theletterf (changes requested):

distro/README.md
MrAlias: Yes, this specifically describes option types that only our distro accepts and the unique defaults we have.

distro/otel.go
MrAlias: This should only ever be reached if the user takes a positive action to set no trace exporter. Even if they make a mistake and set an invalid value the SDK will still use the default exporter. I'm open to suggestions of what we should recommend, but I anticipate users will get here because they want to disable tracing. This log message is only intended to confirm they did so.

theletterf: Got it. How about something along the lines of...

```suggestion
		c.Logger.V(1).Info("OTEL_TRACES_EXPORTER is empty or invalid: Tracing disabled")
```

(If that is indeed the var they'd use and the cases where the condition would be triggered.)

MrAlias: You can get here by passing `OTEL_TRACES_EXPORTER=none` or passing `WithTracesExporter(nil)` directly as an option. If `OTEL_TRACES_EXPORTER` is invalid that is also logged as an error and the default OTLP is used instead.

MrAlias: Also, probably worth mentioning for completeness, if `OTEL_TRACES_EXPORTER` is empty and no `WithTracesExporter` option is passed the default OTLP exporter is also used.

theletterf: Thanks! I've updated my comment to reflect that.

theletterf (approved):

pellared (commented):
In general looks very nice. I left a few comments, please answer them. I will continue with the review tomorrow, but I do not think I will spot anything blocking this PR.

distro/config.go
```diff
@@ -104,6 +116,71 @@ func newConfig(opts ...Option) *config {
...
+		return zapcore.Level(0)
+	case "error":
+		return zapcore.Level(1)
+	default:
+		// "info" or unrecognized level, use info level regardless.
+		return zapcore.Level(-1)
+	}
+}
+
+func zapConfig() *zap.Config {
```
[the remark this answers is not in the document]
MrAlias: The default logging includes a timestamp which would make the output non-deterministic.

distro/config_test.go
```diff
@@ -174,23 +185,93 @@ func TestLoadPropagatorComposite(t *testing.T) {
...
+		t.Run(tc.in, func(t *testing.T) {
+			assert.Equal(t, zapcore.Level(tc.want), zapLevel(tc.in))
+		})
+	}
+}
+
+func TestLoggerPanic(t *testing.T) {
+	zc := zapConfig()
+	// Set an invalid level so the zap logger build will error. This error
+	// shoule be panic-ed.
```
pellared, line 274:
```suggestion
	// should be panic-ed.
```

distro/config.go
```diff
@@ -104,6 +115,87 @@ func newConfig(opts ...Option) *config {
...
+func logger(zc *zap.Config) logr.Logger {
+	z, err := zc.Build()
+	if err != nil {
+		// This should never happen because we control zc. Panic to expose the
+		// bug ASAP to the developer.
+		panic(err)
+	}
+	return zapr.NewLogger(z)
+}
+
```
pellared, line 198: how about moving this to `zap_logger.go` or `default_logger.go` or even `logger.go` ?

distro/config.go
```diff
@@ -43,6 +48,9 @@ const (
 	otelExporterOTLPEndpointKey       = "OTEL_EXPORTER_OTLP_ENDPOINT"
 	otelExporterOTLPTracesEndpointKey = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"
 
+	// Logging level to set.
```
pellared, line 51:
```suggestion
	// Logging level to set when using the default logger.
```

distro/otel.go
```diff
@@ -79,6 +88,12 @@ func Run(opts ...Option) (SDK, error) {
 	if err != nil {
 		return SDK{}, err
 	}
+	if !serviceNameDefined(res) {
+		c.Logger.Info(
+			"service.name attribute is not set. Your service is unnamed and might be difficult to identify. " +
+				`Set your service name using the OTEL_SERVICE_NAME environment variable. For example, OTEL_SERVICE_NAME="<YOUR_SERVICE_NAME_HERE>")`,
+		)
```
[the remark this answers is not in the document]
MrAlias: Verbosity level 0 is a warning. This will be logged as a warning.

pellared (approved):
:shipit:

PR #477, opened 2022-02-21
Use jackc/pgx/v4 consistently.

cloneable:


cloneable:
This would need more work. Some test cases are failing due to changed assumptions around connection config defaults, so some decisions have to be made around this that I cannot do.

pellared:
> This would need more work. Some test cases are failing due to changed assumptions around connection config defaults, so some decisions have to be made around this that I cannot do.

Anyway, I would like to thank you for your effort for trying to fix it 🙇 👍

PR #478, opened 2022-02-22
Use jackc/pgx/v4 consistently

MrAlias:
Fix #477


pellared (approved):
I think it only misses `go mod tidy`. Besides LGTM 👍

PR #492, opened 2022-03-14
Make distro its own module

MrAlias:
Isolate the dependencies of the distro package from all other packages that import splunk-otel-go (e.g. SDK, gRPC, protobuf)


go.mod
```diff
@@ -3,21 +3,7 @@ module github.com/signalfx/splunk-otel-go
...
 require (
-	github.com/go-logr/logr v1.2.2
-	github.com/go-logr/zapr v1.2.3
 	github.com/stretchr/testify v1.7.0
-	github.com/tonglil/buflogr v0.0.0-20220114010534-d490b3990d7e
-	go.opentelemetry.io/contrib/propagators/aws v1.4.0
-	go.opentelemetry.io/contrib/propagators/b3 v1.4.0
-	go.opentelemetry.io/contrib/propagators/jaeger v1.4.0
-	go.opentelemetry.io/contrib/propagators/ot v1.4.0
 	go.opentelemetry.io/otel v1.4.1
```
pellared, line 7: if we want to be strict on isolating the dependencies then we should also create a module for `github.com/signalfx/splunk-otel-go/instrumentation/internal`

see:

```sh
$ go mod why go.opentelemetry.io/otel
# go.opentelemetry.io/otel
github.com/signalfx/splunk-otel-go/instrumentation/internal
go.opentelemetry.io/otel
```

pellared (approved):

go.mod
MrAlias: Yeah, I was thinking the same. I'll approach that in a follow on.

PR #583, opened 2022-03-17
distro: Refactor exporterConfig

pellared:
### Proposed changes

- Remove the OTLP exporter env var handling as OTel Go SDK handles it already in the same way (has the same default)
- Remove `UseTLS` as it was not necessary.


distro/exporter.go
```diff
@@ -41,24 +41,11 @@ var exporters = map[string]traceExporterFunc{
...
-	if c.Endpoint == "" {
-		if endpoint := func() string {
-			// Allow the exporter to use environment variables if set.
-			if _, ok := os.LookupEnv(otelExporterOTLPEndpointKey); ok {
-				return ""
-			}
-			if _, ok := os.LookupEnv(otelExporterOTLPTracesEndpointKey); ok {
-				return ""
-			}
-			return defaultOTLPEndpoint
```
MrAlias, line 53: Removing this is not going to work. If we do not check this and use our default value but the user set either environment variable, those environment variables will be ignored. This comes from the fact that OTel-Go will override environment variables with any passed argument.

MrAlias (changes requested):

distro/exporter.go
pellared: > OTel-Go will override environment variables with any passed argument.

Is it not considered a bug? _Or maybe I understand something bad._ I think no other SDK works like that. Also, I would not come up with this after reading https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/otlp/otlptrace#environment-variables

MrAlias: > Is it not considered a bug?

Not by all. I agree it is not idea, but it has become the released behavior.

> Also, I would not come up with this after reading

Why not? What do you take away from this?

> [Configuration using options have precedence over the environment variables.](https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/otlp/otlptrace#environment-variables)

For completeness:

https://github.com/open-telemetry/opentelemetry-go/blob/a4ea45c38cbcf76c43ab0555b2d4bcacbbfdce9a/exporters/otlp/otlptrace/internal/otlpconfig/options.go#L101-L104

pellared: I will take another look at it tomorrow. Right now, I do not see a problem in my proposed change...

An example scenario when my proposed change would not work will hopefully enlighten me. Then I could debug it and understand it 😄

MrAlias: Is it?! I thought the Jaeger endpoint was different. Let me check.

MrAlias: err, yeah, this is for the OTLP exporter. Yeah, you are correct that is the same.

MrAlias (approved):

PR #723, opened 2022-03-29
Add Splunk span limit defaults

MrAlias:
Conform with the [GDI specification](https://github.com/signalfx/gdi-specification/blob/33f1b65633c5f83acc1622bbc79f498e513a6c7c/specification/configuration.md#instrumentation-libraries):

- Distribution MUST default to 1000 for OTEL_SPAN_LINK_COUNT_LIMIT (not OpenTelemetry default)
- Distribution MUST default to 12000 for OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT (not OpenTelemetry default)
- Distribution MUST be unset "" (unlimited) for all others (not OpenTelemetry default)

### DIFF
--- a/distro/config_test.go
+++ b/distro/config_test.go
@@ -25,6 +25,7 @@ import (
 	"go.opentelemetry.io/contrib/propagators/jaeger"
 	"go.opentelemetry.io/contrib/propagators/ot"
 	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/trace"
 )
 
 type KeyValue struct {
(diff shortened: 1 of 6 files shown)


distro/limits.go
```diff
@@ -0,0 +1,151 @@
...
+	// to zero, which would happen if we delared with &trace.SpanLimits{...}.
+	limits := trace.NewSpanLimits()
+
+	// limits will use OTel defaults or the applicable environment variable if
+	// they are set. The Splunk defaults need to be applied only if the
+	// environment variables were unset.
+	limits = sources.applyDefaults(limits)
+
+	return &limits
+}
```
pellared, line 151: It took me a while to understand this code and to be certain that is OK 😉 

I propose to simplify it to something like this:

```golang
func newSpanLimits() *trace.SpanLimits {
	// Use trace.NewSpanLimits here to ensure any future additions are not set
	// to zero, which would happen if we delared with &trace.SpanLimits{...}.
	limits := trace.NewSpanLimits()

	applyDefault(&limits.AttributeValueLengthLimit, 12000, attributeValueLengthKey, spanAttributeValueLengthKey)
	applyDefault(&limits.AttributeCountLimit, -1, attributeCountKey, spanAttributeCountKey)
	applyDefault(&limits.EventCountLimit, -1, spanEventCountKey)
	applyDefault(&limits.AttributePerEventCountLimit, -1, spanEventAttributeCountKey)
	applyDefault(&limits.LinkCountLimit, 1000, spanLinkCountKey)
	applyDefault(&limits.AttributePerLinkCountLimit, -1, spanLinkAttributeCountKey)

	return &limits
}

// applyDefault sets val to dft if none of envVars  is defined.
func applyDefault(val *int, dft int, envVars ...string){
	for _, env := range envVars {
		if _, ok := os.LookupEnv(env); ok {
			return
		}
	}
	*val = dft
}
```

theletterf (approved):

distro/limits.go
MrAlias: I prefer the current implementation. It encapsulates related values with the type system instead of parameters to a function call, it has methods related to a type defined for that type, uses the language to loop instead of writing each iteration in code, declares the associated limit sources as a package level variable instead of containing them in a function, and doesn't use functions to modify a passed argument instead of using a return value.

That said, I care more about progressing this than keeping with the current approach. I can update if it will achieve this goal, with the caveat that `applyDefaults` will be updated to return a value instead of modifying a passed argument.

MrAlias: Updated.

pellared: > declares the associated limit sources as a package level variable

Side note: I dislike global variables and I avoid them as much as possible 😉 I will probably always complain when I see one - it is not personal 😄   

> I care more about progressing

👍 

> with the caveat that applyDefaults will be updated to return a value instead of modifying a passed argument.

That is a good one.

distro/config_test.go
```diff
@@ -118,6 +119,48 @@ var ConfigurationTests = []*ConfigFieldTest{
...
+			return c.SpanLimits
+		},
+		DefaultValue: &trace.SpanLimits{
+			AttributeValueLengthLimit:   12000,
+			AttributeCountLimit:         -1,
+			EventCountLimit:             -1,
+			LinkCountLimit:              1000,
+			AttributePerEventCountLimit: -1,
+			AttributePerLinkCountLimit:  -1,
+		},
```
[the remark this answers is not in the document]
pellared: Good idea.

pellared (approved):
It would be good to mention the env vars in the markdown docs. But it can be done later. Approving and merging.

PR #725, opened 2022-03-30
Use Splunk realm for endpoint if defined

MrAlias:


theletterf (approved):
Approved. `SPLUNK_REALM` should be added to the settings docs though.

```  
* - ``SPLUNK_REALM``
     - The name of your organization's realm, for example, ``us0``. When you set the realm, traces are sent to ``https://ingest.<realm>.signalfx.com/v2/trace``.
     - ```

File: distro/exporter.go [hunk 1]
```diff
@@ -61,21 +62,33 @@ func newOTLPExporter(c *exporterConfig) (trace.SpanExporter, error) {
...
+	// Allow the exporter to interpret these environment variables directly.
+	envs := []string{otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey}
+	for _, env := range envs {
+		if _, ok := os.LookupEnv(env); ok {
+			return ""
+		}
+	}
+
+	// Use the realm only if OTEL_EXPORTER_OTLP*_ENDPOINT are not defined.
+	if realm, ok := os.LookupEnv(splunkRealmKey); ok {
```
pellared, line 79: Empty (`""`) and `"none"` env var values should be interpreted as if the realm is not defined.

The same comment applies to `func jaegerEndpoint`

References:
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#parsing-empty-value
- https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries

pellared (changes requested):

distro/exporter.go
MrAlias: Updated. PTAL

pellared (approved):

PR #792, opened 2022-04-07
Use TLS if connecting outside of localhost

MrAlias:


distro/exporter.go
```diff
@@ -62,26 +66,52 @@ func newOTLPExporter(c *exporterConfig) (trace.SpanExporter, error) {
...
+		return configured, isLocalhost(configured)
 	}
 
 	// Allow the exporter to interpret these environment variables directly.
 	envs := []string{otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey}
 	for _, env := range envs {
-		if _, ok := os.LookupEnv(env); ok {
-			return ""
+		if v, ok := os.LookupEnv(env); ok {
+			return "", isLocalhost(v)
```
pellared, line 78: I think that it can be changed to:

```suggestion
			return "", false
```

As the OTel Go should be already setting secure/insecure

pellared: We need to test the behavior when the local collector has TLS configured.

MrAlias: Passing the TLS certificate pair you want to test with the `WithTLSConfig`. If the collector is running locally with TLS it will likely be using a non-host CA which means this options will be required regardless of this parsing. If they have updated the localhost CA, that can be accommodated by passing an empty config to that option as well.

pellared: > If the collector is running locally with TLS it will likely be using a non-host CA which means this options will be required regardless of this parsing.


You are right 👍 

And what about "it will not connect to a remote collector that works without TLS." ? E.g. in the case of a K8s or docker-compose env like this

```
version: "3.9"
services:

  // some instrumented Go service
  server:
    // ...
    environment:
      - OTEL_SERVICE_NAME=server

      - OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4317
    depends_on:
      - otel-collector

  otel-collector:
    image: quay.io/signalfx/splunk-otel-collector:0.48.0
    environment:
      - SPLUNK_ACCESS_TOKEN
      - SPLUNK_REALM
    ports:
      - "1777:1777"   # pprof extension
      - "8888:8888"   # Prometheus metrics exposed by the collector
      - "13133:13133" # health_check extension
      - "55679:55679" # zpages extension
      - "4317:4317"   # OTLP gRPC receiver
      - "4318:4318"   # OTLP HTTP receiver
 ```

moreover also similar problem would be when an OTel Collector would be used as a gateway/proxy and one would like to send the telemetry directly to it (without having a collector on localhost).

For sure this scenario should be supported: https://docs.splunk.com/Observability/gdi/get-data-in/application/go/instrumentation/instrument-go-application.html#deploy-the-go-instrumentation-in-kubernetes

pellared: bump 😉

MrAlias: I think I have addressed the issue based on our conversation. Please take another look, and unresolve if not.

pellared:
## Testing

| scenario | trace |
| - | - |
| local collector, default config | https://app.signalfx.com/o11y/#/apm/traces/294ac1e454292882fd392e6f9bb7a94a |

pellared (approved):
## Test Report - gRPC OTLP exporter

| Scenario | Trace |
| - | - |
| default config | https://app.signalfx.com/o11y/#/apm/traces/294ac1e454292882fd392e6f9bb7a94a |
| `SPLUNK_REALM=us0` | https://app.signalfx.com/o11y/#/apm/traces/1f1b805eaa93dbb8fd4dc0816a0f733d |
| `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317` | https://app.signalfx.com/o11y/#/apm/traces/d5e8617cbcc812ee6249c46326e9726f |
| `OTEL_EXPORTER_OTLP_ENDPOINT=http://<Windows_IP>:4317` | https://app.signalfx.com/o11y/#/apm/traces/8bf2ef0b61b4939490a5c41b29ea6f31 |
| `OTEL_EXPORTER_OTLP_ENDPOINT=http://<WSL_IP>:4317` | https://app.signalfx.com/o11y/#/apm/traces/cba1329c95c075dfa27b8b18763a825c  |
| `distro.WithEndpoint("localhost:4317")` | https://app.signalfx.com/o11y/#/apm/traces/1a1505dca694d9c21f8fd2797f20286e  |
| `distro.WithEndpoint("<Windows_IP>:4317")` | spams error as expected | 
| `OTEL_TRACES_EXPORTER=jaeger-thrift-splunk SPLUNK_REALM=us0` | https://app.signalfx.com/o11y/#/apm/traces/2dba7b40df9fdc88603edf9008b31949 |
| `distro.WithEndpoint("<Windows_IP>:4317"), distro.WithTLSConfig(&tls.Config{InsecureSkipVerify: true})` | ⚠️  spams error as well (more info below) | 


⚠️ Right now, I think it is not possible to set an insecure, non-loopback OTLP endpoint in code. It can be done only using env vars. Using `distro.WithTLSConfig(&tls.Config{InsecureSkipVerify: true})` is not working. gRPC Go requires to use `grpc.WithInsecure` or [`insecure.NewCredentials`](https://pkg.go.dev/google.golang.org/grpc@v1.46.0/credentials/insecure#NewCredentials).

Personally, I think it is **not** a problem that has to be addressed in this PR. I imagine nobody will try to use it anyway in the upcoming days/weeks. We can discuss how to address it internally. Probably we can consider adding/changing/removing config options.

## Test Report - Jaeger thrift exporter

| Scenario | Trace |
| - | - |
| `OTEL_TRACES_EXPORTER=jaeger-thrift-splunk SPLUNK_REALM=us0` | https://app.signalfx.com/o11y/#/apm/traces/2dba7b40df9fdc88603edf9008b31949 |
| `OTEL_TRACES_EXPORTER=jaeger-thrift-splunk OTEL_EXPORTER_JAEGER_ENDPOINT=http://localhost:14268/api/traces` | https://app.signalfx.com/o11y/#/apm/traces/38537f6b8d0ff0a724537b075fc662d4 |

I pushed db10bc2 with little changes in `example` which helped me in testing.

## Additional comments 

Regarding this PR I think that:

- You can consider changing the PR description (as probably it will be also the commit message). Maybe something like  `Improve OTLP exporter's TLS configuration` or `Fix OTLP exporter's TLS configuration`?
- Update `CHANGELOG.md`

Approving. Nice work 👍

PR #947, opened 2022-05-25
[WIP] Tidy splunkhttp

pellared:


instrumentation/net/http/splunkhttp/config.go
```diff
@@ -24,18 +24,6 @@ const (
 	envVarTraceResponseHeaderEnabled = "SPLUNK_TRACE_RESPONSE_HEADER_ENABLED" // Adds `Server-Timing` header to HTTP responses
 )
 
-// WithTraceResponseHeader enables or disables the TraceResponseHeaderMiddleware.
-//
-// The default is to enable the TraceResponseHeaderMiddleware if this option is not passed.
-// Additionally, the SPLUNK_TRACE_RESPONSE_HEADER_ENABLED environment variable
-// can be set to TRUE or FALSE to specify this option. This option value will be
-// given precedence if both it and the environment variable are set.
-func WithTraceResponseHeader(v bool) Option {
```
MrAlias, line 33: Removing this removes all the `Option`s from the package. We should remove the `Option` type as it will effectively be an unused part of the API.

pellared: I was thinking of it as well. The problem is that if we would like to add some option function in the future then it will be a breaking change of the `NewHandler` function signature. But maybe we will never need an option type here... What is your preference?

pellared: Without `opt ...Options` this line:

https://github.com/signalfx/splunk-otel-go/blob/d30655a501c919c1d6379ca0655e0d23aeead23d/example/main.go#L61

Could become:

```go
router.Use(splunkhttp.NewHandler)
```

Then maybe we should rename `NewHandler` to `TraceResponseHeaderMiddleware` if we would ever want to add other functionalities to `splunkhttp`?

pellared: TL;DR; summary after offline discussion: Simply remove functional options as they probably won't be needed.

pellared: Addressed 5e54a4c

PR #1172, opened 2022-07-12
Drop support for Go 1.16

pellared:
Drop support for Go 1.16 as `go.opentelemetry.io` did the same.


MrAlias (commented):
~Don't `go.mod` files need to be updated? Like [this](https://github.com/signalfx/splunk-otel-go/blob/984f2cc746cc61fc7ca1d566a6859dc05449fb99/distro/go.mod#L3)?~

Gah, sorry, never mind. GitHub was only showing some of the commits. :facepalm:

MrAlias (approved):

PR #1177, opened 2022-07-18
Bump markdownlintcli to v0.32.0

pellared:
## Why

https://github.com/igorshubovych/markdownlint-cli/releases/tag/v0.32.0

## What

Bump markdownlintcli to v0.32.0

### DIFF
--- a/build/pipeline.go
+++ b/build/pipeline.go
@@ -168,7 +168,7 @@ func taskMarkdownLint(skipDocker goyek.RegisteredBoolParam) goyek.Task {
 				tf.Skip("skipping as Docker is needed")
 			}
 
-			if err := tf.Cmd("docker", "run", "--rm", "-v", WorkDir(tf)+":/workdir", "ghcr.io/igorshubovych/markdownlint-cli:v0.31.1", "**/*.md").Run(); err != nil {
+			if err := tf.Cmd("docker", "run", "--rm", "-v", WorkDir(tf)+":/workdir", "ghcr.io/igorshubovych/markdownlint-cli:v0.32.0", "**/*.md").Run(); err != nil {
 				tf.Error(err)
 			}
 		},


pellared:
@MrAlias PTAL (sorry that the PR initially was having a failing build)

build/pipeline.go
```diff
@@ -168,7 +168,7 @@ func taskMarkdownLint(skipDocker goyek.RegisteredBoolParam) goyek.Task {
 				tf.Skip("skipping as Docker is needed")
 			}
 
-			if err := tf.Cmd("docker", "run", "--rm", "-v", WorkDir(tf)+":/workdir", "ghcr.io/igorshubovych/markdownlint-cli:v0.31.1", "**/*.md").Run(); err != nil {
+			if err := tf.Cmd("docker", "run", "--rm", "-v", WorkDir(tf)+":/workdir", "ghcr.io/igorshubovych/markdownlint-cli:v0.32.0", "**/*.md").Run(); err != nil {
```
pellared, line 171: Do we want to extract the version to some `const` e.g. in `build/versions.go`?

MrAlias, line 171: It would be ideal if we could find a way to leverage dependabot to maintain this dependency instead of having to manually do this.

pellared: https://github.com/signalfx/splunk-otel-go/issues/1179

pellared:
> This PR seems to have surpassed initial scope it was created for. Ideally PRs should be scoped to do one thing and one thing only.
> 
> Given how small it is I'm okay with merging this as is, but we should try to refrain from such PRs going forward.
> 
> Please update the commit message to reflect what is being merged when this is merged.

It also fixes the issues reported by the new version of markdownlint. Nothing more than that 😉

MrAlias:
> > This PR seems to have surpassed initial scope it was created for. Ideally PRs should be scoped to do one thing and one thing only.
> > Given how small it is I'm okay with merging this as is, but we should try to refrain from such PRs going forward.
> > Please update the commit message to reflect what is being merged when this is merged.
> 
> It also fixes the issues reported by the new version of markdownlint. Nothing more than that wink

Oh, gotcha! That makes more sense :facepalm:

PR #1185, opened 2022-07-28
Add cspell linter

pellared:
## Why

[cspell](https://github.com/streetsidesoftware/cspell) linter is used with success in https://github.com/open-telemetry/opentelemetry-dotnet and https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation

We already use [misspell](https://github.com/client9/misspell) in this repo. Therefore at first glance, it looks like adding `cspell` does not add significant value. The difference is that `misspell` works more or less like a "blacklist" and detects common misspellings and `cspell` works like a regular spellchecker (allow list) and relies on having dictionaries.

Should we keep only one of them? Or use both?

In this PR `misspell` is used first to detect misspellings and then `cspell` to find unknown words. The benefit of `misspell` is that it suggests the fix. The benefit of `cspell` linter is that it is more strict. I think we should keep both.

## What

Use `cspell` linter (after `misspell`) in the build pipeline.


pellared:
@theletterf WDYT?

pellared:
I see not interest. Closing

Issue #1373, opened 2022-09-27
Let the logger accept case insensitive values


theletterf:
Hi there,

Most of our distributions are taking `DEBUG` as a value for `OTEL_LAMBDA_LOG_LEVEL` and other log level settings. 

It's currently case sensitive in https://github.com/signalfx/splunk-otel-go/blob/main/distro/logger.go#L36-L60

Let's make it case insensitive: I can see scenarios where devs could use either form. It'd also align with other distros and their docs.

PR #1374, opened 2022-09-27
OTEL_LOG_LEVEL accepts case insensitive values

pellared:
Fixes https://github.com/signalfx/splunk-otel-go/issues/1373

### DIFF
--- a/distro/logger.go
+++ b/distro/logger.go
@@ -52,8 +52,9 @@ var (
 
 // zapLevel returns the parsed zapcore.Level.
 func zapLevel(level string) zapcore.Level {
+	level = strings.ToLower(level)
 	for _, l := range logLevels {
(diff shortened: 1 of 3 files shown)


theletterf (approved):

distro/logger.go
```diff
@@ -51,7 +53,7 @@ var (
 // zapLevel returns the parsed zapcore.Level.
 func zapLevel(level string) zapcore.Level {
 	for _, l := range logLevels {
-		if l.String() == level {
+		if l.String() == strings.ToLower(level) {
```
MrAlias, line 56: No need to call this every loop iteration.

```suggestion
	level = strings.ToLower(level)
	for _, l := range logLevels {
		if l.String() == level {
```

MrAlias (approved):

PR #1378, opened 2022-10-04
Do not link to internal docs in root README.md

pellared:
We should refer only to the official docs 😉


theletterf (approved):
Super approved!

PR #1463, opened 2022-10-17
Fix example module name

MrAlias:


pellared (approved):
Out of curiosity: is this causing any problems?

MrAlias:
> Out of curiosity: is this causing any problems?

PR #1570, opened 2022-11-02
Bump all dependencies and remove Go 1.17 support

pellared:
The dependencies were bumped using `go get -u -t ./...` in all modules except `/build`.

The main build pipeline is executed on Go 1.18 (we could probably make a "matrix" and run it on Go 1.19 as well). I would prefer to do it in a seperate PR.

Support for Go 1.17 is removed.


.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/checkout@v3.1.0
       - uses: actions/setup-go@v3.3.1
         with:
-          go-version: '1.17'
+          go-version: '1.18'
```
pellared, line 40: Maybe we could simply use `1.19` here? We have `compatibility-test` to check test against `1.18`.

MrAlias: SGTM

MrAlias (approved):

.github/workflows/ci.yml
pellared: https://github.com/signalfx/splunk-otel-go/pull/1570/commits/c36d9177d2f6bd60da4746944d4601b5534f324a

PR #1573, opened 2022-11-03
Add a task for bumping dependenies

pellared:


pellared:
> Why not use the approach from https://github.com/open-telemetry/opentelemetry-go/blob/main/.github/workflows/scripts/dependabot-pr.sh?

1. Not coupled to `dependabot` config and `gh` CLI tool
2. Bumps pseudo-versions (not handled by dependabot)
3. Bumps transitive dependencies. These are dependencies as well. Why not bumping them if the tests are passing? It also mitigates the risk if the direct dependencies are abandoned or not well maintained. MAYBE it is handled by dependabot - I do not remember at the moment...
4. I think it is simpler and easier to understand 😉

MrAlias (approved):

PR #1754, opened 2023-02-06
Enable sqlclosecheck linter

pellared:
## Why

Per https://github.com/ryanrolds/sqlclosecheck/releases/tag/v0.4.0

## What

- Enable https://github.com/ryanrolds/sqlclosecheck
- Address the reported issues (close `*sql.Rows` and `*sql.Stmt`)
- Use `assert` instead of `require` in `t.Cleanup` to make sure that that everything is cleaned up (`require` calls `t.Fatal/FailNow` which calls `runtime.Goexit`). As far as I remember the problem would be only if multiple `require` are called in a single callback passed to `t.Cleanup`, but I prefer to use `assert` consistently to not burden my mind each time I see `require` together with `t.Cleanup`. Example [here](https://github.com/signalfx/splunk-otel-go/pull/1754/files#r1098531496)


pellared:
@MrAlias I think that the lint failures in `integration_test.go` files makes sense. In `suite_test.go` I think we could add `//nolint:` for the whole package. WDYT? If you agree I can work on address the linting issues.

MrAlias:
> @MrAlias I think that the lint failures in `integration_test.go` files make sense. In `suite_test.go` I think we could add `//nolint:` for the whole package. WDYT? If you agree I can work on address the linting issues.

Why not just fix the errors by closing the appropriate values?

pellared:
> Why not just fix the errors by closing the appropriate values?

Works for me.

pellared:
@MrAlias PTAL

instrumentation/github.com/go-sql-driver/mysql/splunkmysql/test/integration_test.go
```diff
@@ -64,8 +64,8 @@ func newFixtures(t *testing.T) (*tracetest.SpanRecorder, *trace.TracerProvider,
 	db, err := splunksql.Open("mysql", dsn, splunksql.WithTracerProvider(tp))
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		require.NoError(t, db.Close())
-		require.NoError(t, tp.Shutdown(context.Background()))
```
pellared, line 68: This one would not be called if the assertion in previous line fails.

MrAlias (approved):

PR #1846, opened 2023-02-23
Add process detector

pellared:
## Why

The process attributes can be very useful. It is already supported by other Splunk distributions.

## What

Add process detector OOTB.

### DIFF
--- a/distro/otel.go
+++ b/distro/otel.go
@@ -41,7 +41,7 @@ import (
 
 var errShutdown = errors.New("SDK shutdown failure")
 
-var distroVerAttr = attribute.String("splunk.distro.version", splunkotel.Version())
+const distroVerAttr = "splunk.distro.version"
 
 const tracesSamplerKey = "OTEL_TRACES_SAMPLER"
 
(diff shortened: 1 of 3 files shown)


theletterf (approved):

distro/otel.go
```diff
@@ -126,6 +123,32 @@ func Run(opts ...Option) (SDK, error) {
 	return sdk, nil
 }
 
+func newResource(ctx context.Context) (*resource.Resource, error) {
+	// SDK's default resource.
+	res := resource.Default()
```
MrAlias, line 128: Default uses a detector:

https://github.com/open-telemetry/opentelemetry-go/blob/2c213bf585b07e28dee346f9096a0b64d3fc46f8/sdk/resource/resource.go#L216-L221

Might as well use [`New`](https://github.com/open-telemetry/opentelemetry-go/blob/2c213bf585b07e28dee346f9096a0b64d3fc46f8/sdk/resource/resource.go#L48) with all the detectors we want so there is only one detection cycle.

MrAlias: The `distroVerAttr` could also be wrapped in a [`StringDetector`](https://github.com/open-telemetry/opentelemetry-go/blob/2c213bf585b07e28dee346f9096a0b64d3fc46f8/sdk/resource/builtin.go#L77)

pellared: > Might as well use [New](https://github.com/open-telemetry/opentelemetry-go/blob/2c213bf585b07e28dee346f9096a0b64d3fc46f8/sdk/resource/resource.go#L48) with all the detectors we want so there is only one detection cycle.

I think that [`defaultServiceNameDetector`](https://github.com/open-telemetry/opentelemetry-go/blob/b177f58e09ca7e470d2971da4deb9dc3eabd81ac/sdk/resource/builtin.go#L49) is not publicly available in any other way than via `resource.Default()`. Am I correct?

Anyway the proposed approach makes sure all SDK's default decorators are in place

pellared: >The `distroVerAttr` could also be wrapped in a [StringDetector](https://github.com/open-telemetry/opentelemetry-go/blob/2c213bf585b07e28dee346f9096a0b64d3fc46f8/sdk/resource/builtin.go#L77)

I will try that.

MrAlias: > I think that [`defaultServiceNameDetector`](https://github.com/open-telemetry/opentelemetry-go/blob/b177f58e09ca7e470d2971da4deb9dc3eabd81ac/sdk/resource/builtin.go#L49) is not publicly available in any other way than via `resource.Default()`. Am I correct?

Oof, that looks correct. That is not ideal, but it makes sense to go with the current approach given this limitation.

pellared: From https://github.com/open-telemetry/opentelemetry-go/blob/2c213bf585b07e28dee346f9096a0b64d3fc46f8/sdk/resource/builtin.go#L74-L79

> The resulting Resource will have the specified schemaURL.

I guess this is the reason why we use `resource.NewSchemaless` for Splunk-specific attributes. Right?

pellared: PTAL 169229d7

PR #1848, opened 2023-02-24
Add runtime metrics instrumentation

pellared:
## Why

The runtime metrics can be very useful. It is already supported by other Splunk distributions.

## What

Add (experimental) runtime metrics instrumentation when metrics are enabled.


theletterf (approved):
Which metrics are collected?

pellared:
> Which metrics are collected?

We are using https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/runtime/v0.40.0/instrumentation/runtime/runtime.go . You can find the currently collected metrics there.

I do not think we should currently document concrete metrics it as

1. the package and the metrics are not stable. for instance I am not sure about the `runtime.uptime` name
2. the Go runtime metrics are even not documented under https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md

distro/otel_test.go
```diff
@@ -611,54 +644,64 @@ func (coll *collector) Start(t *testing.T) {
...
-	defer timeout.Stop()
-	select {
-	case got := <-coll.metricsService.requests:
-		return &got
-	case <-timeout.C:
-		return nil
-	}
+func (coll *collector) ExportedMetrics() *metricsExportRequest {
+	// stop to make sure all requests are processed and synchronized
+	coll.grpcSrv.GracefulStop()
```
MrAlias, line 658: This means the collector becomes unusable when this or `ExportedTraces` is called. This seems like a sharp edge that will cause issues in future testing.

Is this graceful stop still needed? If it is do we need to update documentation somewhere to try and help preventing this from causing future issues?

distro/otel_test.go
```diff
@@ -611,54 +644,64 @@ func (coll *collector) Start(t *testing.T) {
...
+			Header:   headers,
+			Resource: rs.GetResource(),
+		}
 	}
+	// concat all spans
+	for _, scopeSpans := range rs.ScopeSpans {
+		ctss.data.Spans = append(ctss.data.Spans, scopeSpans.GetSpans()...)
+	}
+
+	ctss.mtx.Unlock()
```
MrAlias, line 683: Why not defer this?

distro/otel_test.go
```diff
@@ -574,30 +611,26 @@ func (coll *collector) Start(t *testing.T) {
...
-	cmpb.RegisterMetricsServiceServer(srv, coll.metricsService)
+	coll.grpcSrv = grpc.NewServer(opts...)
+	ctpb.RegisterTraceServiceServer(coll.grpcSrv, coll.traceService)
+	cmpb.RegisterMetricsServiceServer(coll.grpcSrv, coll.metricsService)
 	errCh := make(chan error, 1)
 
 	// Serve and then stop during cleanup.
 	t.Cleanup(func() {
-		srv.GracefulStop()
+		coll.grpcSrv.GracefulStop()
```
MrAlias, line 630: This seems redundant with the `GracefulStop` in the telemetry retrieval methods.

distro/otel_test.go
```diff
@@ -611,54 +644,64 @@ func (coll *collector) Start(t *testing.T) {
...
+		cmss.data = &metricsExportRequest{
+			Header:   headers,
+			Resource: rs.GetResource(),
+		}
+	}
+	// concat all metrics
+	for _, scopeMetrics := range rs.ScopeMetrics {
+		cmss.data.Metrics = append(cmss.data.Metrics, scopeMetrics.GetMetrics()...)
 	}
+	cmss.mtx.Unlock()
```
MrAlias, line 704:
```suggestion
	cmss.mtx.Lock()
	defer cmss.mtx.Unlock()
	if cmss.data == nil {
		// headers and resource should be the same. set them once
		cmss.data = &metricsExportRequest{
			Header:   headers,
			Resource: rs.GetResource(),
		}
	}
	// concat all metrics
	for _, scopeMetrics := range rs.ScopeMetrics {
		cmss.data.Metrics = append(cmss.data.Metrics, scopeMetrics.GetMetrics()...)
	}
```

distro/otel_test.go
pellared: It should not be needed. I will double-check it and remove.

pellared: after removal

```
$ go test -race -count=20
PASS
ok      github.com/signalfx/splunk-otel-go/distro       17.476s
```

and after reading the metrics SDK code I agree that it is not needed 😉

distro/otel_test.go
pellared: part of https://github.com/signalfx/splunk-otel-go/pull/1848/commits/854a374398ef4a11b52fcaf03a1f9d6bfb9a63d1

distro/otel_test.go
pellared: not longer valid after https://github.com/signalfx/splunk-otel-go/pull/1848/commits/854a374398ef4a11b52fcaf03a1f9d6bfb9a63d1

distro/otel_test.go
pellared: part of https://github.com/signalfx/splunk-otel-go/pull/1848/commits/854a374398ef4a11b52fcaf03a1f9d6bfb9a63d1

distro/otel_test.go
pellared: part of https://github.com/signalfx/splunk-otel-go/pull/1848/commits/854a374398ef4a11b52fcaf03a1f9d6bfb9a63d1

pellared:
```
$ go test -race -count=20
PASS
ok      github.com/signalfx/splunk-otel-go/distro       17.476s
```

MrAlias (approved):

PR #1973, opened 2023-03-28
Add metrics in splunksql

pellared:
## Example of exported metrics

```json
[
  {
    "Name": "db.client.connections.usage",
    "Description": "The number of connections that are currently in state described by the state attribute",
    "Unit": "{connection}",
    "Data": {
      "DataPoints": [
        {
          "Attributes": [
            {
              "Key": "pool.name",
              "Value": {
                "Type": "STRING",
                "Value": "mockDB://bob@localhost:8080/testDB"
              }
            },
            {
              "Key": "state",
              "Value": {
                "Type": "STRING",
                "Value": "used"
              }
            }
          ],
          "StartTime": "2023-03-28T15:48:28.9293792+02:00",
          "Time": "2023-03-28T15:48:28.9293792+02:00",
          "Value": 0
        },
        {
          "Attributes": [
            {
              "Key": "pool.name",
              "Value": {
                "Type": "STRING",
                "Value": "mockDB://bob@localhost:8080/testDB"
              }
            },
            {
              "Key": "state",
              "Value": {
                "Type": "STRING",
                "Value": "idle"
              }
            }
          ],
          "StartTime": "2023-03-28T15:48:28.9293792+02:00",
          "Time": "2023-03-28T15:48:28.9293792+02:00",
          "Value": 1
        }
      ],
      "Temporality": "CumulativeTemporality",
      "IsMonotonic": false
    }
  },
  {
    "Name": "db.client.connections.max",
    "Description": "The maximum number of open connections allowed",
    "Unit": "{connection}",
    "Data": {
      "DataPoints": [
        {
          "Attributes": [
            {
              "Key": "pool.name",
              "Value": {
                "Type": "STRING",
                "Value": "mockDB://bob@localhost:8080/testDB"
              }
            }
          ],
          "StartTime": "2023-03-28T15:48:28.9293792+02:00",
          "Time": "2023-03-28T15:48:28.9293792+02:00",
          "Value": 0
        }
      ],
      "Temporality": "CumulativeTemporality",
      "IsMonotonic": false
    }
  },
  {
    "Name": "db.client.connections.wait_time",
    "Description": "The time it took to obtain an open connection from the pool",
    "Unit": "ms",
    "Data": {
      "DataPoints": [
        {
          "Attributes": [
            {
              "Key": "pool.name",
              "Value": {
                "Type": "STRING",
                "Value": "mockDB://bob@localhost:8080/testDB"
              }
            }
          ],
          "StartTime": "2023-03-28T15:48:28.9293792+02:00",
          "Time": "2023-03-28T15:48:28.9293792+02:00",
          "Value": 0
        }
      ],
      "Temporality": "CumulativeTemporality",
      "IsMonotonic": false
    }
  }
]
```


instrumentation/database/sql/splunksql/sql.go
```diff
@@ -68,25 +71,57 @@ func Open(driverName, dataSourceName string, opts ...Option) (*sql.DB, error) {
...
+	if poolName == "" {
+		poolName = driverName
+	}
+	reg, err := registerMetrics(db, cfg.ResolveMeter(), poolName)
+	if err != nil {
+		// We prefer reporting problems with metrics rather than failing.
+		otel.Handle(err)
+	}
+	if reg != nil {
+		closerConn.SetMetricsRegistration(reg)
```
pellared, line 121: I need to add a comment why we do this.

instrumentation/database/sql/splunksql/sql.go
```diff
@@ -68,25 +71,57 @@ func Open(driverName, dataSourceName string, opts ...Option) (*sql.DB, error) {
...
+		connCfg, err = regCfg.DSNParser(dataSourceName)
+		if err != nil {
+			otel.Handle(err)
+		}
+	} else {
+		// Fallback. This is a best effort attempt if we do not know how to
+		// explicitly parse the DSN.
+		connCfg, _ = urlDSNParse(dataSourceName)
+	}
+	regOpt := withRegistrationConfig(&connCfg, regCfg)
```
pellared, line 86: I think this should be refactored (at least extracted to a function)

instrumentation/internal/sqltestutil/db_metrics.go
pellared, line 1: I am not sure if `instrumentation/internal/sqltestutil` is a good package. It might be too granular and such approach may lead to "module explosion".

I am also considering to make a `internal/testutil` module as it might contain reusable code for all `*/test` modules. But this on the other hand may lead to a "big ball of mud".

`internal/testutil` could also contain some helpers that would wrap common patterns like:

```go
	reader := metric.NewManualReader()
	meterProvider := metric.NewMeterProvider(metric.WithReader(reader))
	defer func() { assert.NoError(t, meterProvider.Shutdown(ctx)) }()
```

and

```go
rm := metricdata.ResourceMetrics{}
	err = reader.Collect(ctx, &rm)
	require.NoError(t, err)
	require.Len(t, rm.ScopeMetrics, 1, "should export metrics")
	got := rm.ScopeMetrics[0]
```

But too much abstractions does not always help the test code. Less code is not always better as the additional layers of indirecation can casue the tests to be less readable.

Therefore, I currently propose to have `instrumentation/internal/sqltestutil` and I wait for feedback before I reuse it in all instrumentations built on top of `splunksql`.

instrumentation/database/sql/splunksql/config.go
```diff
@@ -121,6 +123,12 @@ func WithTracerProvider(tp trace.TracerProvider) Option {
 	return optionConv{iOpt: internal.WithTracerProvider(tp)}
 }
 
+// WithMeterProvider returns an Option that sets the MeterProvider used with
+// this instrumentation library.
+func WithMeterProvider(mp metric.MeterProvider) Option {
```
MrAlias, line 128: I'm hesitant to add this to the exported types given this package is stably release.

instrumentation/database/sql/splunksql/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package splunksql provides functions to trace the database/sql package
+// (https://golang.org/pkg/database/sql) using the OpenTelemetry API. It will
+// automatically augment operations such as connections, statements and
+// transactions with tracing.
```
MrAlias, line 18:
```suggestion
```

instrumentation/internal/sqltestutil/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+Package sqltestutil contains helpers for testing database instrumentations
+build on top of the splunksql instrumentation.
+This package is in a separate module from the instrumentation it tests to
+isolate the dependency of the default SDK and not impose this as a transitive
+dependency for users.
```
MrAlias, line 20: This package doesn't seem to make sense. The `github.com/signalfx/splunk-otel-go/instrumentation/database/sql/splunksql/test` and `github.com/signalfx/splunk-otel-go/instrumentation/github.com/go-sql-driver/mysql/splunkmysql/test
` packages are already their own modules with isolated dependencies. By sharing this internal module across module boundaries it effectively makes this API public. This is something already done with the config, but I'm not sure we want to continue adding to this. Especially if we are not sure about the design of the API yet (give it will have to be a stable API).

instrumentation/internal/sqltestutil/db_metrics.go
```diff
@@ -0,0 +1,122 @@
...
+								attribute.String("state", "used"),
+							),
+							Value: 0,
+						},
+						{
+							Attributes: attribute.NewSet(
+								wantPoolAttr,
+								attribute.String("state", "idle"),
+							),
+							Value: 1,
```
MrAlias, line 76: How do we know these values will always be `0` and `1` (and, say, not the other way around)?

MrAlias (commented):
Overall the approach of this PR looks good. I'm happy to see the design of `RegisterCallbacks` working well here.

There were some issues though:

- The config rename could be split into another PR to make this PR more targeted.
- The `sqltestutil` package really shouldn't exist.

instrumentation/internal/sqltestutil/doc.go
pellared: > give it will have to be a stable API.

It will not have to be as it is only used in `test` modules. The `instrumentation/internal` module has to be stable. But ``instrumentation/internal/sqltestutil` would not have to be if we use it only in `test` modules.

instrumentation/internal/sqltestutil/db_metrics.go
pellared: If the SQL statement is closed then the pooled connection is idle.

instrumentation/database/sql/splunksql/config.go
pellared: I agree. I will address it.

pellared: The other possibility is to wait with merging until OTel metrics API is stable.

pellared: Agreed to unexport it.

instrumentation/internal/sqltestutil/doc.go
pellared: https://github.com/signalfx/splunk-otel-go/pull/1973/commits/f359c10e63ba2527a92ef0ea6346665311682e45

instrumentation/internal/sqltestutil/db_metrics.go
pellared: https://github.com/signalfx/splunk-otel-go/pull/1973/commits/f359c10e63ba2527a92ef0ea6346665311682e45

instrumentation/database/sql/splunksql/config.go
pellared: I will not have access in `test` module to unexported fields...

I think we can just wait until GA of the metrics API. @MrAlias, what do you think?

pellared: Addressed by:

- e9aebc48
- https://github.com/signalfx/splunk-otel-go/issues/1976

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Add metrics in the following database insturmentation libraries. (#1973)
```
pellared, line 13: Should we call out (e.g. by names) the metrics which are added?

@theletterf @MrAlias

instrumentation/database/sql/splunksql/sql.go
```diff
@@ -12,10 +12,10 @@
...
 // limitations under the License.
 
-// Package splunksql provides functions to trace the database/sql package
-// (https://golang.org/pkg/database/sql) using the OpenTelemetry API. It will
-// automatically augment operations such as connections, statements and
-// transactions with tracing.
+// Package splunksql provides functions to instrument the [database/sql] package
+// using the OpenTelemetry API. It will augment operations
+// such as connections, statements and transactions with tracing
+// and collect connection pool metrics.
```
pellared, line 18: Should we describe the metrics? E.g. like here: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/runtime/doc.go?

Issue #1988, opened 2023-04-03
Bad instrumentation scope version


pellared:
The instrumentation library should contain its own version value instead of reporting the version of the main/distro module.

PR #1990, opened 2023-04-05
Fix distro and instrumentation version values

pellared:
Fixes https://github.com/signalfx/splunk-otel-go/issues/1988

- Fix `telemetry.sdk.version` resource attribute to properly return the `github.com/signalfx/splunk-otel-go/distro` Go module version.
- Fix instrumentation scope version value to properly return the instrumentation library versions.


theletterf (approved):

instrumentation/database/sql/splunksql/test/suite_test.go
```diff
@@ -307,6 +308,7 @@ func (s *SplunkSQLSuite) assertSpans(name moniker.Span, count int, c traceapi.Sp
 			}
 		}
 		s.Equalf(traceapi.SpanKindClient, roSpan.SpanKind(), "span %q is not a client span", name)
+		s.Equal(splunkotel.Version(), roSpan.InstrumentationLibrary().Version, "version should match")
```
MrAlias, line 311: Shouldn't this test against the version from `instrumentation/database/sql/splunksql`?

And similar for the other `test` packages?

MrAlias: I think we can just export `Version`. That is what we do upstream when we need to reference the version like this.

I would refrain from adding a version to the test packages as they are only intended to test their respective instrumentation.

pellared: How about addressing it in a separate PR?  I can create an issue for tracking. I plan to solve it on Tuesday.

pellared: The good thing about doing it in a separate PR is that this PR will be a fix and the next one would expand the API.

pellared: https://github.com/signalfx/splunk-otel-go/issues/1992

MrAlias (approved):

PR #2259, opened 2023-05-24
distro: OTEL_METRICS_EXPORTER defaults to otlp

pellared:


distro/otel_test.go
```diff
@@ -143,6 +143,9 @@ func TestMain(m *testing.M) {
 }
 
 func TestRunJaegerExporter(t *testing.T) {
+	// Do not use the default metrics exporter.
+	t.Setenv("OTEL_METRICS_EXPORTER", "none")
```
pellared, line 147: This is needed as otherwise the `sdk.Shutdown` fails as it cannot export the runtime metrics.

I am considering adding the following (pseudo-)code 

```go
	cleanup := setenv("OTEL_TRACES_EXPORTER", "none") 
	defer cleanup()
	cleanup = setenv("OTEL_METRICS_EXPORTER", "none") 
	defer cleanup()
```

Then the tests would explicitly e.g. `t.Setenv("OTEL_TRACES_EXPORTER", "")` when a default exporter is needed in test.

Something like this https://github.com/signalfx/splunk-otel-go/blob/64f4431794100cd5230e51535705865df007e2bf/distro/helpers_test.go would be brought back 😉

 I would rather do this in a separate PR to make reviewing easier.

pellared: https://github.com/signalfx/splunk-otel-go/issues/2260

distro/otel_test.go
```diff
@@ -413,17 +424,49 @@ func TestRunOTLPMetricsExporterTLS(t *testing.T) {
...
 	coll.Start(t)
-	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "https://"+coll.Endpoint)
+	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://"+coll.Endpoint)
+
+	emitMetric(t)
+
+	got := coll.ExportedMetrics()
+	require.NotNil(t, got)
+	assert.Equal(t, []string{"application/grpc"}, got.Header.Get("Content-type"))
+	assertHasMetric(t, got, metricName)
```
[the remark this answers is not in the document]
pellared: https://github.com/signalfx/splunk-otel-go/issues/2261

PR #2729, opened 2024-01-09
Bump dependencies

pellared:


MrAlias:
Can you list what this is bumping in the description so we have a record of it in the commit history when this is merged?

pellared:
> If it's not too hard, can you list what this is bumping in the description so we have a record of it in the commit history when this is merged?

I updated the PR description and edited the commit message.

PR #3002, opened 2024-02-28
Drop support for Go 1.20

MrAlias:
Support for this version of Go is dropped upstream in the next release.


pellared:
I would rather merge it before bumping to newer OTel Go version as we may need to create a bugfix.

pellared (approved):

PR #3207, opened 2024-04-26
Use renovate for /build

pellared:
From https://docs.renovatebot.com/configuration-options/#packagerules:

> Renovate evaluates all packageRules and does not stop after the first match. Order your packageRules so the least important rules are at the top, and the most important rules at the bottom. This way important rules override settings from earlier rules if needed.

I prefer to keep dependabot config for `/build` before we have confidence that renovate works as expected.


renovate.json
```diff
@@ -15,6 +12,12 @@
       "matchManagers": ["gomod"],
       "matchDepTypes": ["indirect"],
       "enabled": true
+    },
+    {
+      "matchFileNames": ["build/**"],
+      "matchManagers": ["gomod"],
+      "matchDepTypes": ["indirect"],
+      "enabled": false
```
MrAlias, line 20: Why false?

pellared: To disable it as it is enabled in previous rule.

MrAlias (approved):

PR #3230, opened 2024-05-06
[chore] Renovate to not ignore any path

pellared:
Renovate currently does not bump the `test` Go modules because of https://github.com/renovatebot/renovate/discussions/12755.

This PR overrides the setting making sure that no path is ignored.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,6 +3,7 @@
   "extends": [
     "config:base"
   ],
+  "ignorePaths": [],
   "separateMajorMinor": true,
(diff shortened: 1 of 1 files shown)


MrAlias:
codecov looks broken in the CI

pellared:
> codecov looks broken in the CI

```
Tokenless has reached GitHub rate limit.
```

This is not related to the change. I think there was some more "global" problem as I see this issue recently to be more frequent.

MrAlias (approved):

PR #3374, opened 2024-08-08
Add Container ID to resource if available

### LINKED ISSUE #3355: Add container ID to resource if running in a detectable environment
When running in an containerized environment it is advantageous to have the container ID present on the resource. This can be used to help classify data and find its origin.

The distro does not currently include this information in its resource. Adding the [`WithContainerID`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/resource#WithContainerID) option to our resource construction will add this information.

The addition of this detector will have minimal overhead. If not running in a containerized environment, it will be a no-op and not affect the produced resource.

MrAlias:
Use the `resource.WithContainer` option to include all available information about the container an application is running in. If a container is not running in a container, this is will not add any attributes and will not alter the existing resource definitions.

This does not add tests due to the challenge in mocking a container runtime environment. It relies on the upstream OTel testing of the `WithContainer` options instead.

Fix #3355


pellared (approved):
How about adding to changelog something like

```
### Added

- Add container attributes to resource if available. (#3374)
```

Issue #3514, opened 2024-11-28
Add opt-in logs support


pellared:
Add opt-in support for setting up a `LoggerProvider` similar to https://github.com/signalfx/splunk-otel-go/pull/1835.

Our `example` can be updated to use [`otelslog`](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelslog).

PR #3673, opened 2025-02-07
distro: Add opt-in logs support

ysolomchenko:
## Why

Fixes https://github.com/signalfx/splunk-otel-go/issues/3514

## What

Add opt-in logs support. The user has to set `OTEL_LOGS_EXPORTER` env var to `otlp` to enable the OTLP gRPC logs exporter (default is `none`).

In a separate PR I will expand `example` with logging support.


CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
```
pellared, line 11:
```suggestion
The release adds logs support.
Set `OTEL_LOGS_EXPORTER=otlp` to enable logs support.
However, take notice that as for now
the OpenTelemetry Go Logs API is not stable.

### Added
```

pellared (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+The release adds logs support.
```
MrAlias, line 11:
```suggestion
This release adds experimental logs support.
```

CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+### Added
+
+- Add logs support and `OTEL_LOGS_EXPORTER` environment variable.
+  `OTEL_LOGS_EXPORTER` accepts:
+  `none` - logs disabled,
+  `otlp` - OTLP gRPC exporter.
+  Currently `OTEL_LOGS_EXPORTER` defaults to `none`
+  as OpenTelemetry Go logs API and SDK are not stable yet.
+  Set `OTEL_LOGS_EXPORTER=otlp` to enable
+  logs support.
```
MrAlias, line 25: Please reformat using the existing style of the document: wrap at an 80 character limit.

CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+Set `OTEL_LOGS_EXPORTER=otlp` to enable logs support.
+However, take notice that as for now
+the OpenTelemetry Go Logs API is not stable.
+
+### Added
+
+- Add logs support and `OTEL_LOGS_EXPORTER` environment variable.
+  `OTEL_LOGS_EXPORTER` accepts:
+  `none` - logs disabled,
+  `otlp` - OTLP gRPC exporter.
```
MrAlias, line 21: This needs to be restructured to use markdown syntax for lists.

distro/exporter.go
```diff
@@ -236,6 +238,51 @@ func newOTLPMetricsExporter(c *exporterConfig) (metric.Exporter, error) {
...
+
+		return logsExporters[defaultLogsExporter]
+	}
+	return lef
+}
+
+func newOTLPlogExporter(c *exporterConfig) (log.Exporter, error) {
+	ctx := context.Background()
+
+	var opts []otlploggrpc.Option
```
MrAlias, line 267: What about HTTP support?

pellared: This is OK.
It is only needed for our trace and metrics ingest endpoint. We do not have logs ingest (sic!).

CHANGELOG.md
```diff
@@ -8,6 +8,20 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+
+### Added
+
+- Added logs support and the `OTEL_LOGS_EXPORTER` environment variable.
+- `OTEL_LOGS_EXPORTER` accepts:
+  - `none` - logs disabled.
+  - `otlp` - OTLP gRPC exporter.
+- Currently, `OTEL_LOGS_EXPORTER` defaults to `none` as the OpenTelemetry Go
+  Logs API and SDK are not stable yet.
+- Set `OTEL_LOGS_EXPORTER=otlp` to enable logs support.
```
pellared, line 23: This should be one entry (not 4).

pellared: https://github.com/signalfx/splunk-otel-go/pull/3673/commits/d7ecd5b5f845ba2a583bc546202dd2358527253b

distro/exporter.go
MrAlias: What about sending logs to a local collector over HTTP?

pellared: @MrAlias 

We are always sending to collector using gRPC for all signals.
See:
- https://github.com/signalfx/splunk-otel-go/blob/46aa0ab7055386411c699579cbf38dd88f345bf3/distro/exporter.go#L60-L92
- https://github.com/signalfx/splunk-otel-go/blob/46aa0ab7055386411c699579cbf38dd88f345bf3/distro/exporter.go#L205-L237

There was never an ask to add support for `OTEL_EXPORTER_OTLP_PROTOCOL`.

Maybe we can consider reusing [`autoexport` ](https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport) module as a separate PR. What do you think?

pellared:
@ysolomchenko, please fix the build

distro/exporter.go
MrAlias: I'm confused. Why do we support HTTP for the other signals?

pellared: Because  ingest  is (or was) not supporting gRPC.

```
// Direct ingest to Splunk Observabilty Cloud using HTTP/protobuf. 
```

MrAlias: > Even if we do support gRPC as well I heard that it is better to use `http/protobuf` with our backend. This is why AFAIK all our distros are using `http/protobuf` when `SPLUNK_REALM` is used. We use `grpc` in other cases for backwards compatibility.

Right ... so shouldn't we support HTTP? This is only adding gRPC exporters but you are saying that HTTP is preferred, right?

pellared: > Right ... so shouldn't we support HTTP? This is only adding gRPC exporters but you are saying that HTTP is preferred, right?

It will break users that use `OTEL_EXPORTER_OTLP_ENDPOINT` which targets gRPC endpoints. 

Reference: https://docs.splunk.com/observability/en/gdi/get-data-in/application/go/configuration/advanced-go-otel-configuration.html

MrAlias (approved):

pellared (changes requested):
Some changes (for sure in `CHANGELOG.md` and a comment in `exporter.go`) are lost when you were rebasing and signing. 

Marking as "request changes" so that we do not accidentally merge this PR until all changes are brought back.

pellared (approved):

pellared:
@ysolomchenko, thank you for your first contribution :medal_sports:

pellared:
Few notices for sake of https://docs.splunk.com/observability/en/gdi/get-data-in/application/go/configuration/advanced-go-otel-configuration.html update:
- logs are disabled by default. user has to set `OTEL_LOGS_EXPORTER=otlp` to enable it so that logs are exported using [OTLP gRPC exporter](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc)
- setting `SPLUNK_REALM` does not affect where logs are send. the user can configure the endpoint by setting the [OTLP gRPC exporter](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc) env vars. the most important ones are `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`
- AFAIK it is not possible to send logs directly to Splunk o11y without using an OTel Collector with HEC exporter

PR #3846, opened 2025-04-22
chore: update ubuntu runners to ubuntu-24.04

pellared:
Per https://github.com/actions/runner-images/issues/11101
ubuntu-20.04 runners are deprecated (and the actions using it can fail right now).
See: https://github.com/actions/runner-images


MrAlias:
The expected checks needs to be updated to not expect `ubuntu-20.04`.

Issue #3923, opened 2025-06-09
Support `OTEL_EXPORTER_OTLP_PROTOCOL` et al


pellared:
Add support for following env vars:
- `OTEL_EXPORTER_OTLP_PROTOCOL`
- `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`
- `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL`
- `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL`

These need to default to `grpc` by default for backwards-compatibility.

pellared:
This code

https://github.com/signalfx/splunk-otel-go/blob/de8277446c5673e446e4036e1c707271249189ea/distro/exporter.go#L77-L93

should setup an `otlptracehttp` exporter if `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` or `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf`.

Similarly for metrics and logs.

ysolomchenko:
I’ll work on this

PR #3930, opened 2025-06-13
Support OTEL_EXPORTER_OTLP_TRACES_PROTOCOL

ysolomchenko:
Towards #3923 

## Changes

Added support for following env vars:

- OTEL_EXPORTER_OTLP_PROTOCOL
- OTEL_EXPORTER_OTLP_TRACES_PROTOCOL


distro/exporter.go
```diff
@@ -36,6 +36,8 @@ import (
 
 type traceExporterFunc func(logr.Logger, *exporterConfig) (trace.SpanExporter, error)
 
+const httpProtobuf = "http/protobuf"
```
pellared, line 39: I propose to move it to `config.go` becasue this is the place were we have all the configuration constants defined.

distro/exporter.go
```diff
@@ -298,3 +305,27 @@ func noneEnvVarSet(envs ...string) bool {
 func notNone(s string) bool {
 	return s != "" && s != "none"
 }
+
+func getOTLPProtocol(l logr.Logger, signalKey string) string {
```
pellared, line 309: in Go you try to avoid `get` in the function/method names. Related to: https://go.dev/doc/effective_go#Getters

```suggestion
func otlpProtocol(l logr.Logger, signalKey string) string {
```

distro/exporter.go
```diff
@@ -59,19 +61,24 @@ func tracesExporter(l logr.Logger) traceExporterFunc {
...
-	splunkEndpoint := otlpRealmTracesEndpoint()
-	if splunkEndpoint != "" {
-		// Direct ingest to Splunk Observabilty Cloud using HTTP/protobuf.
-		return otlptracehttp.New(ctx,
-			otlptracehttp.WithEndpoint(splunkEndpoint),
-			otlptracehttp.WithURLPath(otlpRealmTracesEndpointPath),
-			otlptracehttp.WithHeaders(map[string]string{
-				"X-Sf-Token": c.AccessToken,
-			}),
-		)
```
pellared, line 74: `SPLUNK_REALM` need to have precedence over `OTEL_EXPORTER_OTLP_PROTOCOL`. Otherwise if people set only `SPLUNK_REALM` then it would not configure the exporters to use the o11y endpoints. This would be a backwards-incomaptible change.

distro/exporter.go
```diff
@@ -298,3 +305,27 @@ func noneEnvVarSet(envs ...string) bool {
 func notNone(s string) bool {
 	return s != "" && s != "none"
 }
+
+func getOTLPProtocol(l logr.Logger, signalKey string) string {
+	// signal-specific key takes precedence
+	if v := os.Getenv(signalKey); v != "" {
+		if v == defaultOTLPProtocol || v == httpProtobuf {
```
pellared, line 312: this would be more readable

```suggestion
		if v == otlpProtocolGRPC || v == otlpProtocolHTTPProtobuf {
```
and

```go
const defaultOTLPProtocol = otlpProtocolGRPC
```

pellared (changes requested):

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Added
+
+- Support for OTLP protocol configuration via environment variables:
+  - `OTEL_EXPORTER_OTLP_PROTOCOL` to specify the protocol used by OTLP exporters.
```
pellared, line 14: The changelog entry does not say that the new functionality works only for the OTLP trace exporters.

distro/exporter.go
ysolomchenko: Fixed in 7d9b5a5

distro/exporter.go
ysolomchenko: Fixed in 7d9b5a5

distro/otel_test.go
```diff
@@ -695,7 +695,7 @@ func emitLogs(t *testing.T, opts ...distro.Option) {
 	require.NoError(t, sdk.Shutdown(ctx))
 }
 
-func assertHasSpan(t *testing.T, got *spansExportRequest) {
+func asssertHasSpan(t *testing.T, got *spansExportRequest) {
```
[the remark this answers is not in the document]
ysolomchenko: This was already there, and as you mentioned in the comment, I’ll fix these letter errors in a separate PR.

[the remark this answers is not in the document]
pellared: My mistake! I was comparing commit diff instead of "all changes".

distro/exporter.go
```diff
@@ -74,6 +74,11 @@ func newOTLPTracesExporter(_ logr.Logger, c *exporterConfig) (trace.SpanExporter
 		)
 	}
 
+	protocol := otlpProtocol(l, otelTracesExporterOTLPProtocolKey)
+	if protocol == otlpProtocolHTTPProtobuf {
+		return otlptracehttp.New(ctx)
```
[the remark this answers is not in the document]
pellared: Make sure to also cover it in the tests :wink:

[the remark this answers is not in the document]
ysolomchenko: Fixed in 3ac604e

Test added in 4d1fe91

distro/exporter_test.go
```diff
@@ -71,6 +73,41 @@ func TestOTLPTracesEndpoint(t *testing.T) {
 	})
 }
 
+func TestGetOTLPProtocol(t *testing.T) {
```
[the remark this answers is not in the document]
ysolomchenko: fixed in c706189

distro/exporter.go
```diff
@@ -74,23 +74,43 @@ func newOTLPTracesExporter(l logr.Logger, c *exporterConfig) (trace.SpanExporter
 		)
 	}
 
+	headers := make(map[string]string)
+	if c.AccessToken != "" {
+		headers["X-Sf-Token"] = c.AccessToken
+	}
+
+	useInsecure := c.TLSConfig == nil &&
+		noneEnvVarSet(otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey, splunkRealmKey)
```
pellared, line 83: 1. I think that the name should tell about the "state"
2. There is no need to check for `c.TLSConfig == nil` as you use it in `else`

How about?

```go
	isNone := noneEnvVarSet(otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey, splunkRealmKey)
```

or even for lazy evaulation

```go
	isNone := func() bool {
		return noneEnvVarSet(otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey, splunkRealmKey)
	}
```

ysolomchenko: Fixed in [b314012](https://github.com/signalfx/splunk-otel-go/pull/3930/commits/b3140123146f7127ec7001394f07e61b62d1e469)

pellared (approved):

distro/exporter.go
```diff
@@ -298,3 +320,28 @@ func noneEnvVarSet(envs ...string) bool {
...
+//nolint:unparam // This will receive other input values in future.
+func otlpProtocol(l logr.Logger, signalKey string) string {
+	// Signal-specific key takes precedence.
+	if v := os.Getenv(signalKey); v != "" {
+		if v == otlpProtocolGRPC || v == otlpProtocolHTTPProtobuf {
+			return v
+		}
+		err := fmt.Errorf("invalid %s: %q", signalKey, v)
+		l.Error(err, "using default %s: %q", signalKey, defaultOTLPProtocol)
+		return defaultOTLPProtocol
```
MrAlias, line 333: Don't we want to fallback to `OTEL_EXPORTER_OTLP_PROTOCOL` if `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` is unparsable before we result to the default?

distro/exporter.go
```diff
@@ -298,3 +320,28 @@ func noneEnvVarSet(envs ...string) bool {
 func notNone(s string) bool {
 	return s != "" && s != "none"
 }
+
+//nolint:unparam // This will receive other input values in future.
+func otlpProtocol(l logr.Logger, signalKey string) string {
+	// Signal-specific key takes precedence.
+	if v := os.Getenv(signalKey); v != "" {
```
MrAlias, line 327: Should we support values like `"gRPC"` as well?

distro/exporter.go
```diff
@@ -74,18 +74,40 @@ func newOTLPTracesExporter(_ logr.Logger, c *exporterConfig) (trace.SpanExporter
...
+
+		if c.TLSConfig != nil {
+			opts = append(opts, otlptracehttp.WithTLSClientConfig(c.TLSConfig))
+		} else if isNone {
+			// Assume that the default endpoint (local collector) is non-TLS.
+			opts = append(opts, otlptracehttp.WithInsecure())
+		}
+
+		return otlptracehttp.New(ctx, opts...)
+	}
```
MrAlias, line 99: This call to `otlpProtocol` is unnecessary if `isNone` is `true` right?

```suggestion
	if !isNone {
		protocol := otlpProtocol(l, otelTracesExporterOTLPProtocolKey)

		if protocol == otlpProtocolHTTPProtobuf {
			var opts []otlptracehttp.Option

			if len(headers) > 0 {
				opts = append(opts, otlptracehttp.WithHeaders(headers))
			}

			if c.TLSConfig != nil {
				opts = append(opts, otlptracehttp.WithTLSClientConfig(c.TLSConfig))
			} else if isNone {
				// Assume that the default endpoint (local collector) is non-TLS.
				opts = append(opts, otlptracehttp.WithInsecure())
			}

			return otlptracehttp.New(ctx, opts...)
		}
	}
```

ysolomchenko: `isNone` is used to check whether the connection should be insecure. I’ve renamed the variable to avoid confusion. The same check was also present in the gRPC example, so I decided to extract it in order to reduce duplicated code.

Renamed in a74aeb9

distro/exporter.go
ysolomchenko: Changed in 0dfedc7
Test updated in 17adcfa

distro/exporter.go
```diff
@@ -74,18 +74,40 @@ func newOTLPTracesExporter(_ logr.Logger, c *exporterConfig) (trace.SpanExporter
 		)
 	}
 
+	headers := make(map[string]string)
+	if c.AccessToken != "" {
+		headers["X-Sf-Token"] = c.AccessToken
+	}
+	shouldUseInsecure := noneEnvVarSet(otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey, splunkRealmKey)
```
pellared, line 81:
```suggestion
	isLocalCollector := noneEnvVarSet(otelExporterOTLPEndpointKey, otelExporterOTLPTracesEndpointKey, splunkRealmKey)
```

I think would be a lot more readable.

distro/exporter.go
pellared: > This call to otlpProtocol is unnecessary if isNone is true right?

I do not think so.
`isNone` refers to endpoint URL and `otlpProtocol` to the protocol being used.
When `isNone` is `true` then still we need to check what is `otlpProtocol` to check whether we should use `http/protobuf` or `grpc` when sending telemetry to the local collector.

distro/exporter.go
```diff
@@ -298,3 +320,26 @@ func noneEnvVarSet(envs ...string) bool {
...
+
+//nolint:unparam // This will receive other input values in future.
+func otlpProtocol(l logr.Logger, signalKey string) string {
+	// Signal-specific key takes precedence.
+	if v := os.Getenv(signalKey); v != "" {
+		if v == otlpProtocolGRPC || v == otlpProtocolHTTPProtobuf {
+			return v
+		}
+		err := fmt.Errorf("invalid %s: %q", signalKey, v)
+		l.Error(err, "falling back to general %q", otelExporterOTLPProtocolKey)
```
pellared, line 332: nit

```suggestion
		l.Error(err, "falling back to %q", otelExporterOTLPProtocolKey)
```

distro/exporter.go
pellared: Fine. I will create an issue for it here an work on the spec level to allow it.

PR #4370, opened 2026-02-17
Remove splunkelastic

pellared:
https://github.com/olivere/elastic has received no updates since Mar 19, 2022 and it is officially deprecated.

The currently supported and official Go client for Elasticsearch has native OTel instrumentation: https://www.elastic.co/docs/reference/elasticsearch/clients/go/advanced/observability

Moreover, Elasticsearch v7 is not supported since April 15, 2025 (https://support.liferay.com/w/elasticsearch-7-17-end-of-life-eol-timeline-and-liferay-dxp-elasticsearch-compatibility-update-faq) which makes it problematic for testing as it looks like the new Docker Engine does not like the container that we were testing against.

We have also no report of anyone using `splunkelastic` instrumentation library.

This also fixes the test failures on CI.
Since Docker Engine v29+ requires an IP (and it does not accept hostnames anymore)
This issue become visible because of https://github.com/actions/runner-images/issues/13474.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Removed
 
 - Remove `github.com/signalfx/splunk-otel-go/instrumentation/gopkg.in/olivere/elastic/splunkelastic`
-  Go module.
+  Go module. Note that the currently supported and official [Go client for Elasticsearch](https://github.com/elastic/go-elasticsearch)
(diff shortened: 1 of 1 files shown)


pellared:
> * This is doing more that what the title describes "Remove splunkelastic". Consider splitting into multiple PRs
> 
>     * This is a stable module. Do we want to deprecate before removing?
> 
>     * Can you ignore these tests while deprecating?

Will do. Closing the PR as it will be easier just to create a new one.

Issue #4395, opened 2026-02-19
Remove splunkchi instrumentation library


pellared:
We plan to remove the splunkchi instrumentation library from the repository.

To avoid introducing a breaking change without notice, we are going first formally deprecate the library and only remove it in a subsequent release.

PR #4406, opened 2026-02-19
Deprecate splunkchi

pellared:
Towards  https://github.com/signalfx/splunk-otel-go/issues/4395


instrumentation/github.com/go-chi/chi/splunkchi/go.mod
```diff
@@ -1,3 +1,5 @@
+// Deprecated: the module is not going to be released in future.
```
pellared:
```suggestion
// Deprecated: this module is no longer supported.
```

PR #4409, opened 2026-02-19
Deprecate splunkhttprouter

pellared:
Towards https://github.com/signalfx/splunk-otel-go/issues/4399


instrumentation/github.com/julienschmidt/httprouter/splunkhttprouter/httprouter.go
```diff
@@ -14,6 +14,9 @@
 
 // Package splunkhttprouter provides OpenTelemetry instrumentation for the
 // github.com/julienschmidt/httprouter module.
+//
+// Deprecated: the module is not going to be released in future.
```
pellared:
```suggestion
// Deprecated: this module is no longer supported.
```

PR #4880, opened 2026-09-18
Align span collection limits with OpenTelemetry defaults

pellared:
Follows https://github.com/signalfx/gdi-specification/pull/414

- Align span attribute, event, link, and per-event/link attribute count limits with OpenTelemetry defaults.
- Preserve environment variable overrides and the Splunk-specific 12,000 attribute value length default.
- Document the breaking behavior change, its data-dropping impact, and settings that preserve the previous defaults.


MrAlias (dismissed):

MrAlias:
This means the next release is going to be v2.0, right?

